PR #1, opened 2025-01-15
Add ASRTTSProvider

openminddev:
Add an ASRTTSProvider that we can use in the ASR and TTS

### DIFF
--- a/src/providers/asr_tts_provider.py
+++ b/src/providers/asr_tts_provider.py
@@ -8,6 +8,9 @@ class ASRTTSProvider:
     ASRTTSProvider is a singleton class that manages the Text-to-Speech (TTS) state.
     Thread-safe implementation ensures consistent state across multiple threads.
     """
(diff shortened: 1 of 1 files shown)


src/runtime/global_state.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        with lock:
+            if cls not in instances:
+                instances[cls] = cls(*args, **kwargs)
+        return instances[cls]
+
+    return get_instance
+
+@singleton
+class GlobalState:
```
openmindchiral: I think it would be more explicit to have a new directory (global) and then have a different class for each type of global state

src/providers
   {name_of_provider}.py
   ...

I encourage this because the interface for state being set is clearer to understand without inspecting the dynamic dict

openminddev: Sure. A clear state is good!

PR #2, opened 2025-01-17
Diagnostic Logging V0

openmindchiral:
This is a bare-minimum PR to demonstrate persistent logging of tick-based logging using SQLite.

As a follow up to this, a webserver that read from SQLite would have to be connected to the datastore to display relevant diagnostic data


src/runtime/cortex.py
```diff
@@ -38,15 +41,19 @@ async def _run_cortex_loop(self) -> None:
             await self._tick()
 
     async def _tick(self) -> None:
+        tick_id = str(uuid.uuid4())
```
openmindchiral, line 44: This associates a unique ID for every "tick" which occurs in the system to tag events and actions against

TangmereCottage:
address through more recent approaches

PR #41, opened 2025-01-29
Support twitter bot

boya-crypto:
add twitter and rag inputs  so that user can send tweets based on his query

### DIFF
--- a/src/inputs/base/__init__.py
+++ b/src/inputs/base/__init__.py
@@ -16,8 +16,9 @@ class SensorOutputConfig:
     **kwargs : dict
         Additional configuration parameters
     """
+
     base_url: T.Optional[str] = None
-    
+
     def __init__(self, base_url: T.Optional[str] = None, **kwargs):
         self.base_url = base_url
         # Store any additional config parameters
--- a/src/inputs/plugins/twitter.py
+++ b/src/inputs/plugins/twitter.py
@@ -17,7 +17,7 @@ class TwitterInput(FuserInput[str]):
         config: Optional[SensorOutputConfig] = None,
     ):
         """Initialize TwitterInput with configuration.
-        
+
         Parameters
         ----------
         config : Optional[SensorOutputConfig]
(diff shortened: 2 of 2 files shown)


src/inputs/plugins/openmind.py
```diff
@@ -0,0 +1,86 @@
...
+    
+    def __init__(self):
+        super().__init__()
+        self.buffer: List[str] = []
+        self.message_buffer: Queue[str] = Queue()
+
+        # Initialize OpenMind provider
+        self.provider = OpenMindProvider()
+        self.provider.start()
+        self.provider.register_message_callback(self._handle_message)
```
openminddev: BTW, we don't need the provider. We can directly do the API requests in the plugin.
The provider is used to make sure that it requires the constant connection for example: webcam, ws connection, etc.

src/run.py
```diff
@@ -1,23 +1,47 @@
...
+    logging.info("Attempting to send first boot tweet...")
+    send_first_boot_tweet(config)
+    
+    if config_name == "twitter" and query:
+        # Run with query
+        runtime = CortexRuntime(config)
+        asyncio.run(runtime.run())
+        return
+        
+    # Normal startup
```
openminddev: Is this necessary?

boya-crypto: fixed, I add query into the config file, we can have same run time flow as other agents

src/inputs/plugins/openmind.py
boya-crypto: removed provider

boya-crypto:
> * Could you please delete `.omos/first_boot_tweet`?
> * could you please rename OpenMindProvider to something else? OpenMindProvider seems confusing to me
> * Please follow the naming conventions for providers.
> * Please fix all tests and lint.

I have fixed the issues mentioned above, the only unit test failure i got is
This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3

src/run.py
```diff
@@ -14,6 +15,7 @@
 def start(config_name: str, debug: bool = False) -> None:
     logging.basicConfig(level=logging.DEBUG if debug else logging.INFO)
     config = load_config(config_name)
+    send_first_boot_tweet(config)
```
openminddev: Please remove this.

openminddev:
> > * Could you please delete `.omos/first_boot_tweet`?
> > * could you please rename OpenMindProvider to something else? OpenMindProvider seems confusing to me
> > * Please follow the naming conventions for providers.
> > * Please fix all tests and lint.
> 
> I have fixed the issues mentioned above, the only unit test failure i got is This request has been automatically failed because it uses a deprecated version of `actions/upload-artifact: v3

Sure! One last minor change.

src/run.py
boya-crypto: I have removed it

PR #56, opened 2025-02-01
Feat/pygame improvements

ahmadkhan100:
feat(simulator): improve pygame implementation and UI
- Add pygame initialization helpers
- Improve thread safety for macOS
- Move Input History to left side panel
- Add ETH balance display
- Enhance UI layout and organization

### DIFF
--- a/src/run.py
+++ b/src/run.py
@@ -4,7 +4,9 @@ import threading
 import platform
 import multiprocessing
 import time
+import queue
 import os
+import pygame
 
 import dotenv
 import typer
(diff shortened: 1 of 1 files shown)


src/run.py
```diff
@@ -10,17 +17,82 @@
 
 app = typer.Typer()
 
+class MacOSSimulatorLoop:
```
openminddev, line 20: Could you please put this logic to the plugins? The main entry file should be clean and easy to understand.

src/simulators/plugins/pygame_init.py
```diff
@@ -0,0 +1,9 @@
+import os
+import pygame
+import platform
+
+def init_pygame():
```
openminddev, line 5: Maybe move this one to RacconSim?

src/simulators/plugins/RacoonSim.py
```diff
@@ -2,108 +2,719 @@
...
 from providers.io_provider import IOProvider
 
 
 class RacoonSim:
+    _instance = None
+    _init_lock = threading.Lock()
+    _event_queue = Queue()
+    _main_thread_initialized = threading.Event()
+
+    def __new__(cls):
```
openminddev: We have a singleton provide, but why do we need a singleton provider?

src/run.py
ahmadkhan100: Alright doing.

src/simulators/plugins/pygame_init.py
ahmadkhan100: Okay

src/simulators/plugins/RacoonSim.py
ahmadkhan100: Let me see, should remove

PR #57, opened 2025-02-01
Fix/pygame macos init

ahmadkhan100:
## Changes
1. RacoonSim.py:
   - Removed singleton pattern as it's not needed
   - Moved pygame initialization directly into RacoonSim
   - Added proper macOS environment variable handling
   - Improved error handling for thread initialization

2. run.py:
   - Kept MacOSSimulatorLoop in run.py for macOS-specific handling
   - Improved platform-specific initialization logic
   - Maintained clear separation between macOS and other platform handling

3. simulator_loop.py:
   - Kept as modular component for potential cross-platform use
   - Provides base simulator loop functionality

### DIFF
--- a/src/run.py
+++ b/src/run.py
@@ -4,7 +4,9 @@ import threading
 import platform
 import multiprocessing
 import time
+import queue
 import os
+import pygame
 
 import dotenv
 import typer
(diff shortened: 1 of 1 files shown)


src/run.py
```diff
@@ -10,17 +17,82 @@
 
 app = typer.Typer()
 
+class MacOSSimulatorLoop:
```
openminddev, line 20: The whole logic should not be here. The workflow of the whole system is reading the configuration json file and loading the right inputs and outputs.

Also, could you please explain why we need another thread for the tick which has already been baked in the https://github.com/OpenmindAGI/omOS/blob/f02237acb719fdd7a0d7c2e4bbd35dcb12bbb151/src/simulators/orchestrator.py#L39 ?

ahmadkhan100: I have removed this. i did not check this. Thanks for informing

PR #58, opened 2025-02-01
Refactor/remove duplicate threading

ahmadkhan100:
# Remove Duplicate Thread Management

## Changes
- Removed redundant `MacOSSimulatorLoop` since threading is handled by orchestrator
- Created `platform_handlers.py` for platform-specific initialization
- Simplified RacoonSim by removing thread management code
- Cleaned up run.py to use orchestrator's thread management

## Why
The orchestrator already provides thread management for simulators in `_run_simulator_loop`, making the additional thread management in run.py unnecessary and potentially problematic.

## Testing
- Verified simulator initialization works correctly
- Confirmed platform-specific settings are properly applied
- Tested simulator loop performance using orchestrator's thread management


src/run.py
```diff
@@ -1,26 +1,34 @@
...
+import multiprocessing
+import platform
 
 import dotenv
 import typer
 
 from runtime.config import load_config
 from runtime.cortex import CortexRuntime
 from runtime.robotics import load_unitree
+from simulators.plugins.platform_handlers import init_platform
```
openminddev: should we put this to the `__enter__` of the racoonsim?

src/run.py
```diff
@@ -1,26 +1,34 @@
...
     config = load_config(config_name)
     _ = load_unitree(config)
     runtime = CortexRuntime(config)
-    asyncio.run(runtime.run())
 
+    # Run normally using orchestrator's thread management
+    asyncio.run(runtime.run())
 
 if __name__ == "__main__":
+    multiprocessing.freeze_support()
```
openminddev: Please remove this

src/simulators/plugins/RacoonSim.py
```diff
@@ -2,108 +2,698 @@
...
+            # Set up frame timing
+            self.clock = pygame.time.Clock()
+            self.frame_count = 0
+            self.fps_update_time = time.time()
+            self.last_frame = time.time()
+            
+            return True
+            
+        except Exception as e:
+            print(f"Error initializing pygame: {str(e)}")
```
openminddev: Please use logging rather than print

src/simulators/plugins/simulator_loop.py
```diff
@@ -0,0 +1,43 @@
+import pygame
+import queue
+
+class SimulatorLoop:
```
openminddev: Do we need to keep this file?

ahmadkhan100: removed

src/run.py
ahmadkhan100: done.

src/run.py
ahmadkhan100: done.

openminddev:
I removed the thread checking. All good now.

PR #59, opened 2025-02-02
Governance

TangmereCottage:
works but gives error message:

"Error fetching blockchain data: ENS name: '' is invalid."

### DIFF
--- a/tests/inputs/base/test_governance_ethereum.py
+++ b/tests/inputs/base/test_governance_ethereum.py
@@ -25,9 +25,11 @@ def mock_requests_post():
     with patch("requests.post") as mock:
         yield mock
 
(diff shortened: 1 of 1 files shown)


openminddev:
works on my end
```
INFO:root:load_rules_from_blockchain()
INFO:root:Blockchain response: 200
INFO:root:Blockchain data: {'rules': 'You are honest and friendly. You would never do anything to hurt a human or an animal. Also, you will not approach a human or animal closer than 50 centimeters.'}
INFO:root:7777 rules: You are honest and friendly. You would never do anything to hurt a human or an animal. Also, you will not approach a human or animal closer than 50 centimeters.
```

PR #66, opened 2025-02-04
upgrade spot with local vision

TangmereCottage:


src/runtime/config.py
```diff
@@ -88,7 +88,9 @@ def load_config(config_name: str) -> RuntimeConfig:
         raw_config = json.load(f)
 
     # Load Unitree robot communication channel
-    load_unitree(raw_config)
+    UNITREE_WIRED_ETHERNET = os.environ.get("UNITREE_WIRED_ETHERNET")
+    if UNITREE_WIRED_ETHERNET and UNITREE_WIRED_ETHERNET != "SIM":
```
openminddev: These two line changes are not necessary; `load_unitree` has already performed the environment checks.

openminddev: Fix

PR #84, opened 2025-02-10
Add configuration support to simulators similar to the base input system

ahmadkhan100:
# Add Configuration Support to Simulators

## Changes
- Added `SimulatorConfig` class similar to `SensorConfig` for handling simulator configurations
- Updated `Simulator` base class to accept config parameter in `__init__`
- Modified simulator loader to handle config loading from JSON
- Updated WebSim to use new config system


openminddev:
LGTM
Could you please actually use the variables from the config file in the Sim?

openminddev (approved):

Issue #83, opened 2025-02-10
Architecture Feedback


TangmereCottage:
1. I wouldn't initialize the dotenv here, unless there's something odd happening with threads?
https://github.com/OpenmindAGI/OM1/blob/6ea4effa8cd7935d57ac9ef238de6ac9b0ca0921/src/actions/tweet/first_boot.py#L13

2. Given all the actions implemented so far are passthroughs, it's a bit verbose to redefine the class every time. [https://github.com/OpenmindAGI/OM1/blob/6ea4effa8cd7935d57ac9ef238de6ac9b0ca0921/src/actions/face/implementation/passthrough.py]

You could bake 'passthrough' into the framework load_action fn, so every action implicitly supports passthrough without writing boilerplate files
https://github.com/OpenmindAGI/OM1/blob/6ea4effa8cd7935d57ac9ef238de6ac9b0ca0921/src/actions/__init__.py#L58-L60C13

openminddev:
Do we want to remove `passthrough`? If we bake passthrough into the `load_action` function, I guess we will no longer use it since it is the default setup.

openminddev:
@boya-crypto please fix the first issue

PR #86, opened 2025-02-11
Reduce passthrough boilerplate

hunter-at-topology:
Adds a dynamic hook in `load_action` to set the implementation to a generic passthrough in actions.passthrough

Removes redundant boilerplate code for passthrough implementations


hunter-at-topology:
Addresses. #83

openminddev:
LGTM

PR #91, opened 2025-02-13
Added rag plugin

moaaz12-web:


openminddev:
- Please run `pre-commit run --all-files` and delete unnecessary comments.
- Please delete data folder and add it to .gitignore.

config/rag.json
```diff
@@ -0,0 +1,36 @@
...
+        }
+      }
+    ],
+    "cortex_llm": {
+      "type": "OpenAILLMv2",
+      "config": {
+        "api_key": "om1_live611277",
+        "doc_directory": "./data/faiss_index",
+        "openai_api_key": "YOUR OPENAI API KEY",
+        "question":"What is the document about and who are the authors of it?"
```
openminddev: The question should be based on the prompt rather than entering by yourself. 
The prompt is generated by system with the question that users ask (for example, ASR).

src/inputs/plugins/rag.py
```diff
@@ -0,0 +1,366 @@
...
+                return None
+
+            result = f"""
+            RagInput CONTEXT
+            // START
+            {content}
+            // END
+            """
+
+    async def start(self):
```
openminddev, line 163: Is this needed?

src/llm/plugins/rag_openai.py
```diff
@@ -0,0 +1,121 @@
...
+
+import openai
+from pydantic import BaseModel
+
+from llm import LLM, LLMConfig
+
+R = T.TypeVar("R", bound=BaseModel)
+
+
+class OpenAILLMv2(LLM[R]):
```
openminddev: Please rename it to OpenAILLMRag

openminddev (commented):
Overall look good. Please clean up your code.

TangmereCottage:
@moaaz12-web amazing work! A few questions:

* We have seen erratic performance with pdfs in particular, compared to text data. Please consider adding support for `.txt` and `.md` files, that certainly makes testing easier too. 
* In terms of program flow, the example you provided - hardcoded question - illustrates how to use the system, but in the general case, the robot will want to RAG specific calls to the core LLM. Imagine for example a person asks the robot "_I'm Jane, what sports do I like_", and then after the ASR step, the fuser (possibly) will want to query the KB, receive relevant data ("Jane likes to swim"), and then, add that to the prompt to the core LLM.
* One simple way for the system to do that might be to:
-- When new data are added to a folder, reindex the KB.
-- Allow different parts of the system to query the KB. For example, imagine there is a KBprovider in `/providers`, that can be queried by the Fuser right after receiving a voice question.

PR #118, opened 2025-02-19
Support OpenAI and Gemini as VLM

kyle-openmind:


openminddev:
Looks good! One minor problem: could you please use `VLM` rather than `Vlm` to maintain naming consistency?

src/providers/vlm_websocket_provider.py
```diff
@@ -10,7 +10,7 @@
 
 
 @singleton
-class VLMProvider:
+class VLMWebSocketProvider:
```
openminddev: Could you please rename `VLMWebSocket` to `VLMVila`? That's our underlying model.

kyle-openmind: Sure

TangmereCottage (approved):

PR #138, opened 2025-03-03
Fix speak action and enhance error handling in LLM history manager

ahmadkhan100:


src/providers/llm_history_manager.py
```diff
@@ -85,51 +88,103 @@ async def summarize_messages(self, messages: List[ChatMessage]) -> ChatMessage:
...
-                        else:
-                            raise Exception("Failed to summarize the state")
+                    elif summary_message.role == "system" and "Error" in summary_message.content:
+                        logging.error(f"Summarization failed: {summary_message.content}")
+                        # Keep existing messages but try to reduce size if too large
+                        if len(messages) > 10:
+                            # Remove oldest messages to keep size manageable
+                            while len(messages) > 5:
+                                messages.pop(0)
+                            logging.info("Kept existing messages but reduced history size")
```
openminddev: These hardcode changes don't look right. It is better to just pop the first two conversations, rather than popping messages until there is only 5 left. This will cause us to lose memory.

src/providers/llm_history_manager.py
```diff
@@ -85,51 +88,103 @@ async def summarize_messages(self, messages: List[ChatMessage]) -> ChatMessage:
...
+                except asyncio.CancelledError:
+                    logging.warning("Summary task callback cancelled")
                 except Exception as e:
-                    logging.error(f"Error in summary task callback: {e}")
-                    messages.pop(0) if messages else None
-                    messages.pop(0) if messages else None
+                    logging.error(f"Error in summary task callback: {type(e).__name__}: {e}")
+                    # Prevent history loss by keeping at least some recent messages
+                    while len(messages) > 5 and len(messages) > 0:
+                        messages.pop(0)
```
openminddev: Same here.

src/providers/llm_history_manager.py
```diff
@@ -85,51 +88,103 @@ async def summarize_messages(self, messages: List[ChatMessage]) -> ChatMessage:
...
+            logging.warning("Summary task creation cancelled")
         except Exception as e:
-            logging.error(f"Error starting summary task: {e}")
+            logging.error(f"Error starting summary task: {type(e).__name__}: {e}")
+            # Add retry mechanism for critical failures
+            if "messages" in str(e).lower() or "history" in str(e).lower():
+                logging.info("Attempting to recover from history error")
+                # Preserve some history in case of error
+                while len(messages) > 3 and len(messages) > 0:
+                    messages.pop(0)
```
openminddev: Same here.

openminddev (changes requested):

src/actions/speak/connector/riva_tts.py
```diff
@@ -2,9 +2,10 @@
 from actions.speak.interface import SpeakInput
 from providers.asr_provider import ASRProvider
 from providers.riva_tts_provider import RivaTTSProvider
+import logging
 
 
-class SpeakRos2Connector(ActionConnector[SpeakInput]):
+class SpeakRivaTTSConnector(ActionConnector[SpeakInput]):
```
openminddev, line 7: Please change the config files for this name change

src/providers/llm_history_manager.py
```diff
@@ -85,51 +88,103 @@ async def summarize_messages(self, messages: List[ChatMessage]) -> ChatMessage:
...
             self._summary_task.add_done_callback(callback)
 
+        except asyncio.CancelledError:
+            logging.warning("Summary task creation cancelled")
         except Exception as e:
-            logging.error(f"Error starting summary task: {e}")
+            logging.error(f"Error starting summary task: {type(e).__name__}: {e}")
+            # Add retry mechanism for critical failures
+            if "messages" in str(e).lower() or "history" in str(e).lower():
+                logging.info("Attempting to recover from history error")
```
openminddev: This `try...except` can be triggered when the task creation is failure like the above `asyncio.CancelledError:`, so I don't think we need to verify the error messages here

openminddev:
LGTM

TangmereCottage (approved):

Issue #111, opened 2025-02-18
Discord integration as input and output


kyle-openmind:
Part 1.

Integration with [Discord API](https://discord.com/developers/docs/resources/message#get-channel-messages) as an input plugin for OM1 so that we can interact with OM1 in a Discord Channel.

Part 2.
Integrate OM1 output as text message in Discord.

PR #192, opened 2025-04-05
Integrate discord bot

Steven-Hsu1:
## Overview
Closes: #111 

## Changes
created two files: discord_bot.py and generate.py where we use OM1 API endpoint to generate responses to discord messages. 

## Testing
Screenshot 2025-04-04 at 10 52 43 PM

Sometimes hallucinates that it's name is Cortex. Could be a prompting issue?

Screenshot 2025-04-04 at 11 03 45 PM

## Impact
Allows users to converse with OM1 agent in a specified channel. Saves conversation history and allows for different agents using the different config files. 

## Additional Information
Need to fill .env with additional information:
DISCORD_TOKEN=
CHANNEL_ID= 
OPENMIND_PROVIDER= # supported LLM provider by openmind (openai, Gemini, etc) default: openai
OPENMIND_MODEL= # specific LLM model default: gpt-4o


bobjiang:
Overall, your implementation didn't integrate in OM1,

Steven-Hsu1:
Ah I see, should've implemented a plugin.

Steven-Hsu1:
Retried the issue. Could you take a look again @bobjiang 

Screenshot 2025-04-14 at 9 31 08 PM

bobjiang:
just tried the latest implementation, 
not bad.

one more request, 
please only response when user @bot in the channel, instead of responding continuously like following:

image

src/inputs/plugins/discord.py
```diff
@@ -0,0 +1,300 @@
...
+from inputs.base import SensorConfig
+from inputs.base.loop import FuserInput
+
+# Try to import the Discord connector registration
+try:
+    from actions.speak.connector.discord_message import register_discord_client
+    HAS_CONNECTOR = True
+except ImportError:
+    HAS_CONNECTOR = False
+    logging.warning("Discord message connector not available, bot won't respond automatically")
```
openminddev: Don't import the function from the action.

src/inputs/plugins/discord.py
```diff
@@ -0,0 +1,300 @@
...
+                    "is_bot": True
+                })
+                
+                logging.info(f"Sent message: {content}")
+                self.print_conversation_log()
+                return True
+            return False
+        except Exception as e:
+            logging.error(f"Error sending message: {str(e)}")
+            return False
```
openminddev: Should this be in the action?

src/inputs/plugins/discord.py
```diff
@@ -0,0 +1,300 @@
...
+            # Handle when receiving dict from passthrough
+            content = text['action']
+        else:
+            # Handle direct string input
+            content = text
+            
+        # Actually send the message
+        success = await self.send_message(content)
+        logging.info(f"Discord message sent: {content}, success: {success}")
+        return success
```
openminddev: Should this be in the action?

src/inputs/plugins/discord.py
```diff
@@ -0,0 +1,350 @@
...
+            
+            # Reset the mention flag after sending a response
+            self.last_message_was_mention = False
+            
+            logging.info(f"Sent message: {content}")
+            self.print_conversation_log()
+            return True
+        except Exception as e:
+            logging.error(f"Error sending message: {str(e)}")
+            return False
```
openminddev: Please don't send messages in the input plugin.

bobjiang:
overall, for the project directories, you can refer to https://docs.openmind.org/getting-started/architecture#project-structure

please don't mess up the dirs

bobjiang:
> overall, for the project directories, you can refer to https://docs.openmind.org/getting-started/architecture#project-structure
> 
> please don't mess up the dirs


I would like to emphasis the **data flow**, 

- messages (questions) in the discord channel (@ discord bot)
- input/plugin/discord-input received messages and buffered
- send above messages to LLM (managed by OM1)
- get the responding messages and send them to actions/discord/your-action.py
- through action.py sending the responses to discord channel.

but from your implementations, all the logics are in "input/plugin/"
please read our architectures carefully and re-try it.

Steven-Hsu1:
I believe the above implementation is correct based on the data flow in the documentation. The only action that is required is sending the message (speak) which is implemented within discord_message.py for which we need 3 things: the channel_id to send the message to, the client to send it from, and the message itself. The way I have structured it is that we have a wrapper for the two portions we need from the input file (client and channel_id) and we pass that over to the action file which is when the code that actually sends the message happens. 

The current workflow is as follows: A user @ bot with some message and we store information on who mentioned it, message contents etc --> We take that message and process it within the input function (DiscordInput Fuser, I split it up into multiple functions) --> OM1 sends that to LLM and comes back with response (through the buffer) --> The action orchestrator then takes the wrapper info (channel_id and client) and is able to send the message it got from OM1 to the intended destination using the send_discord_message function. Let me know if this clarifies the process and any other questions you have.

Steven-Hsu1:
I believe the above implementation is correct based on the data flow in the documentation. The only action that is required is sending the message (speak) which is implemented within discord_message.py for which we need 3 things: the channel_id to send the message to, the client to send it from, and the message itself. The way I have structured it is that we have a wrapper for the two portions we need from the input file (client and channel_id) and we pass that over to the action file which is when the code that actually sends the message happens.

The current workflow is as follows: A user @ bot with some message and we store information on who mentioned it, message contents etc --> We take that message and process it within the input function (DiscordInput Fuser, I split it up into multiple functions) and setup the bot as well with proper events --> OM1 sends that to LLM and comes back with response (through the buffer) --> The action orchestrator then takes the wrapper info (channel_id and client) and is able to send the message it got from OM1 to the intended destination using the send_discord_message function. Let me know if this clarifies the process and any other questions you have.

EDIT: I guess the only deviation is that I am registering the bot within the input file from which I am taking that function from the action controller (_register_with_connector) so it has the provided details. How would you like me to handle this? I can reverse the logic so I would import from the input file and grab the needed elements but this seems counter-intuitive.

Issue #239, opened 2025-05-13
Feature: Play sound every minute to keep bluetooth speakers awake


TangmereCottage:
Bluetooth speakers tend to turn themselves off after X minutes to save battery. This is annoying when a robot tries to speak after a long silence, only to find that the bluetooth speaker is asleep. 

Proposed solution - whenever a speaker is connected to the system, play a soft sound every 60 seconds, to keep the system warm.  

Related to that, bluetooth speakers tend to clip the first second (and therefore the first word) of a multi second voice output. Potential solution - add several hundred ms of quiet audio to every audio output, which will then be clipped by the bluetooth out, preventing the first word to be garbled.

ahmadkhan100:
this is done @TangmereCottage in OM1-Modules, please test it

PR #249, opened 2025-05-26
initial commit for feature #239

ahmadkhan100:
Feature #239


ahmadkhan100:
@kyle-openmind can you check why the integration tests are failing?

ahmadkhan100:
> @kyle-openmind can you check why the integration tests are failing?

nw it works now

openminddev:
Could you please add the logic to audio_output_stream.py in the OM1-Modules? This way, we won’t need an extra provider to handle the configuration—it can be a default feature.
1. Play a soft sound every 60 minutes. The sound should be very short to avoid conflicts with actual audio.
2. Add blank words to the base64-encoded audio_data or play a short blank audio before the actual audio.

PR #254, opened 2025-05-27
Add API key docs

openminddev:
## Overview
Add api key docs

### DIFF
--- a/docs/api-reference/openapi.yml
+++ b/docs/api-reference/openapi.yml
@@ -357,6 +357,143 @@ paths:
         - apiKeyAuth: []
         - bearerAuth: []
 
+  /api_keys:
+    get:
+      summary: List API keys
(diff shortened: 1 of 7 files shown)


docs/api-reference/openapi.yml
```diff
@@ -357,6 +357,143 @@ paths:
...
+                      description: The prefix of the API key.
+                    created_at:
+                      type: string
+                      format: date-time
+                      description: The timestamp when the API key was created.
+                    revoked_at:
+                      type: string
+                      format: date-time
+                      description: The timestamp when the API key was revoked, if applicable.
+                    deleted:
```
bobjiang, line 397: if an API key is deleted, 
do we still store the api keys?

openminddev: Yes

PR #277, opened 2025-06-09
better cli interaction

ahmadkhan100:
just do openmind setup keys and run config from there


openminddev:
This is Python project. Installing extra nodejs doesn't make sense. A better way is to figure out how to use Python cli to interact with it.

PR #284, opened 2025-06-11
Add unitree go2 state background check

openminddev:
## Overview
Add unitree Go2 state background task and trigger the `balanceStand` if the mode is in the `jointLock`. The `jointLock` can be triggered when you move the robot to another place. In the `jointLock` mode, it blocks the none zero movement command.

This also improves the movement detection by using the go2 state to check whether the robot is moving or not.

```
Note:
More tests are needed.
```


ahmadkhan100:
bugbot run

PR #433, opened 2025-10-07
Replace the json structure output with function calls

openminddev:
This PR replaces the json structure output with function calls to produce the result more accurate.


pendorre:
love u openmnd

PR #476, opened 2025-10-20
Fix motor_speed setter validation in RPLidar driver

sohw400:
## Problem

The `motor_speed` setter was using `assert` for input validation (line 185). This causes issues in production environments when Python is run with the `-O` flag, which disables assertions. Invalid PWM values would pass through validation and potentially damage the hardware or cause unexpected behavior.

## Solution

Replaced the assertion with explicit `ValueError` exception that:
- Always executes regardless of Python optimization flags
- Provides clear error message with actual and expected values
- Follows Python best practices for input validation

## Changes

- Updated `motor_speed` setter to use `if not (0 <= pwm <= MAX_MOTOR_PWM)` check
- Raises `ValueError` with descriptive message when PWM is out of valid range [0, 1023]

## Testing

The validation logic correctly:
- Accepts PWM values from 0 to 1023
- Rejects negative values with clear error message
- Rejects values above 1023 with clear error message

### DIFF
--- a/src/providers/rplidar_driver.py
+++ b/src/providers/rplidar_driver.py
@@ -182,7 +182,10 @@ class RPDriver(object):
 
     @motor_speed.setter
     def motor_speed(self, pwm):
-        assert 0 <= pwm <= MAX_MOTOR_PWM
+        if not (0 <= pwm <= MAX_MOTOR_PWM):
+            raise ValueError(
+                f"PWM value must be between 0 and {MAX_MOTOR_PWM}, got {pwm}"
+            )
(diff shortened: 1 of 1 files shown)


openminddev:
Hi @sohw400, this looks great! Could you please create a separate PR to only include the changes you mentioned for this? Thanks!

sohw400:
Done! I've created a separate PR with only the RPLidar motor_speed validation fix:

https://github.com/OpenMind/OM1/pull/478

This PR now focuses solely on replacing the `assert` with proper `ValueError` exception handling.

PR #477, opened 2025-10-20
Introduce lifecycle

openminddev:
## Overview
This PR introduces the lifecycle for multiple modes where users can trigger the lifecycle when switching between different modes

### DIFF
--- a/src/actions/move_go2_autonomy/connector/unitree_sdk_advance.py
+++ b/src/actions/move_go2_autonomy/connector/unitree_sdk_advance.py
@@ -9,10 +9,10 @@ import zenoh
 
 from actions.base import ActionConfig, ActionConnector, MoveCommand
 from actions.move_go2_autonomy.interface import MoveInput
+from providers.face_presence_provider import FacePresenceProvider
 from providers.odom_provider import OdomProvider, RobotState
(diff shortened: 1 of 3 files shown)


sohw400:
Found a small inconsistency in the `on_failure` default values. The `LifecycleHook` dataclass defaults to `"ignore"` (line 49), but `parse_lifecycle_hooks` uses `"log"` as the default (line 344). This could cause unexpected behavior depending on how the hooks are created. Should these be aligned?

openminddev:
> Found a small inconsistency in the `on_failure` default values. The `LifecycleHook` dataclass defaults to `"ignore"` (line 49), but `parse_lifecycle_hooks` uses `"log"` as the default (line 344). This could cause unexpected behavior depending on how the hooks are created. Should these be aligned?

Yes, good catch!

openminddev:
> Found a small inconsistency in the `on_failure` default values. The `LifecycleHook` dataclass defaults to `"ignore"` (line 49), but `parse_lifecycle_hooks` uses `"log"` as the default (line 344). This could cause unexpected behavior depending on how the hooks are created. Should these be aligned?

Could you please send me an email at boyuan@openmind.org? We’d love to chat about some interesting opportunities.

PR #621, opened 2025-10-29
perf(websim): merge server thread into uvicorn event loop for lower latency

savataqe1:
## Overview
This PR improves the performance and stability of the WebSim realtime state updates by ensuring all WebSocket communication occurs inside a single, consistent asyncio event loop.
The previous multi-loop implementation added unnecessary latency and introduced risks of runtime loop-mismatch errors.

## Changes

- Unified WebSim runtime under uvicorn’s event loop 
- Removed blocking time.sleep(0.5) from tick()
- Replaced loop.run_until_complete() with thread-safe scheduling:
- asyncio.run_coroutine_threadsafe(self.broadcast_state(), self._server_loop)
- Ensured WebSocket sends happen only on the server’s loop
- Minor stability improvements when removing disconnected clients
- Added safer initialization for _server_loop to avoid tooling warnings
- Result: broadcasting is now event-driven, lower latency, less CPU consumption

## Impact
Performance,Reliability,Behavior impact:
Marked as a safe improvement, but technically timing-related → potentially breaking if external code relied on old delay

### DIFF
--- a/src/simulators/plugins/WebSim.py
+++ b/src/simulators/plugins/WebSim.py
@@ -495,12 +495,12 @@ class WebSim(Simulator):
     def _run_server(self):
         """Run the FastAPI server"""
         # Create and own an event loop in this server thread and expose it
-        self._server_loop: Optional[asyncio.AbstractEventLoop] = None
+        self._server_loop: asyncio.AbstractEventLoop = asyncio.get_event_loop()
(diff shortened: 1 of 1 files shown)


savataqe1:
@openminddev  @boyuan-chen  @Prachi1615

PR #623, opened 2025-10-29
Feat/update to recent face

WenjinFu:
## Overview
Adding the Suppression rule to the face presence provider:
If (frames_recent >= min_obs_window) AND (frames_with_unknown / frames_recent < unknown_frac_threshold),
          then suppress unknowns (report 0). Otherwise, report the **peak**
          number of unknown faces observed in any single frame within the window.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ logs/
 
 # Mode system state files
 config/memory
+config/.runtime.json5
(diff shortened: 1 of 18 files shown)


src/providers/face_presence_provider.py
```diff
@@ -253,12 +291,5 @@ def _fetch_snapshot(self, recent_sec: Optional[float] = None) -> PresenceSnapsho
...
 
-        self._unknown_faces = unknown
-
         ts = float(data.get("server_ts", time.time()))
         return PresenceSnapshot(ts=ts, names=names, unknown=unknown, raw=data)
-
-    @property
-    def unknown_faces(self):
-        """Return the most recent count of unknown faces detected."""
-        return self._unknown_faces
```
openminddev, line 264: Could you please resort this? We need this for the action.

openminddev (commented):
Please fix the lint

PR #674, opened 2025-11-08
Update function_schemas.py

cryptoahmet06:
Handles missing input type hints in function schema generation by logging a warning and returning an empty schema, preventing runtime errors and improving robustness.


cryptoahmet06:
I've run ruff check locally and applied all fixes. Also ran black and isort to ensure formatting and import order are correct. CI lint failures are due to unrelated missing files/star-import warnings, everything else is fixed and ready for review.

PR #694, opened 2025-11-11
feat: integrate avatar commands for thinking in llm

YuchengZhou821:
**Avatar state management:**

* When a request is sent to the LLM, the avatar is set to "Think" to indicate processing.
* After processing, if no face-related action is present in the response, the avatar is set to "Happy" to signal completion.
* The `AvatarProvider` is newly imported to support these avatar updates.


openminddev:
Please fix the lint and add unit tests for it

PR #710, opened 2025-11-13
Update unitree_sdk.py

cryptoahmet06:
Summary

Added a safe initialization block for AudioClient in EmotionUnitreeConnector.
Prevents crashes when initialization fails and adds clear error logging.

Changes

Wrapped AudioClient.Init() in try/except

Added status check for initialization return code

Ensures LED initialization only runs on success

Impact

Improves stability of Unitree emotion connector during startup.

### DIFF
--- a/config/unitree_go2_modes.json5
+++ b/config/unitree_go2_modes.json5
@@ -133,6 +133,11 @@
       },
       "agent_actions": [
         {
+          "name": "face",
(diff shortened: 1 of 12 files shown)


openminddev:
We have moved the Gazebo to `unitree-sdk`. Thanks.

PR #729, opened 2025-11-17
add response message

YuchengZhou821:
This pull request introduces health check functionality to the `AvatarProvider` by adding a publisher and subscriber for avatar health check messages and updating the message definitions to support request-response communication. The main changes are grouped into Zenoh communication improvements and message structure updates.

**Zenoh communication improvements:**

* Added `avatar_healthcheck_publisher` and `avatar_subscriber` to `AvatarProvider`, enabling the provider to respond to health check requests via a new `_handle_avatar_request` method. This sets up a subscriber on `"om/avatar/request"` and a publisher on `"om/avatar/response"` for health check responses.
* Updated the initialization logic to declare both publishers and the subscriber, and improved logging to reflect the new topics.

**Message structure updates:**

* Modified `AvatarFaceRequest` and `AvatarFaceResponse` dataclasses to include a `request_id` field, enabling request-response correlation. Removed the `code` field from `AvatarFaceResponse` and replaced it with `request_id`.
* Updated imports in `avatar_provider.py` to include `AvatarFaceResponse` for health check response handling.
* When sending avatar commands, now explicitly sets `request_id` to an empty string in the `AvatarFaceRequest`.

### DIFF
--- a/tests/integration/test_case_runner.py
+++ b/tests/integration/test_case_runner.py
@@ -5,6 +5,7 @@ import re
 import time
 from pathlib import Path
 from typing import Any, Dict, List, Tuple
+from unittest.mock import MagicMock, patch
 
 import json5
 import openai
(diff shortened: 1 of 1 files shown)


src/providers/avatar_provider.py
```diff
@@ -18,27 +23,57 @@ def __init__(self):
...
             )
+            self.running = True
+            logging.info("AvatarProvider initialized with Zenoh on topics")
         except Exception as e:
             logging.error(f"Failed to initialize AvatarProvider Zenoh session: {e}")
-            self.session = None
-            self.avatar_publisher = None
-            self.running = False
+
+    def _handle_avatar_request(self, sample):
```
openminddev: Please fix the data type and the docstring

src/providers/avatar_provider.py
```diff
@@ -18,27 +23,57 @@ def __init__(self):
...
+        """
+        Handle incoming avatar requests.
+
+        Processes health check requests and responds.
+        """
+        try:
+            request = AvatarFaceRequest.deserialize(sample.payload.to_bytes())
+
+            if request.code == AvatarFaceRequest.Code.ACTIVE.value:
```
openminddev: This looks like wrong. Do we want to listen to the status?

src/providers/avatar_provider.py
```diff
@@ -66,6 +101,8 @@ def send_avatar_command(self, command: str) -> bool:
 
             face_msg = AvatarFaceRequest(
                 header=prepare_header(str(uuid4())),
+                request_id="",
+                code=AvatarFaceRequest.Code.UNKNOWN.value,
```
openminddev: Why is this `UNKNOWN`?

src/zenoh_msgs/idl/status_msgs.py
```diff
@@ -121,12 +121,25 @@ class ASRText(IdlStruct, typename="ASRText"):
 
 @dataclass
 class AvatarFaceRequest(IdlStruct, typename="AvatarFaceRequest"):
+    class Code(Enum):
+        ACTIVE = 0
+        INACTIVE = 1
+        UNKNOWN = 2
```
openminddev: Is the code correct?

src/providers/avatar_provider.py
```diff
@@ -66,6 +114,8 @@ def send_avatar_command(self, command: str) -> bool:
 
             face_msg = AvatarFaceRequest(
                 header=prepare_header(str(uuid4())),
+                request_id="",
```
openminddev: Should we use str(uuid4())?

src/providers/avatar_provider.py
```diff
@@ -16,30 +23,71 @@ class AvatarProvider:
...
         self.avatar_publisher = None
+        # Health Check Publisher and Subscriber
+        self.avatar_healthcheck_publisher = None
+        self.avatar_subscriber = None
         self.running = False
 
-        self._initialize_zenoh()
-
-    def _initialize_zenoh(self):
+    def start(self):
```
openminddev: Should we use the original method?

src/providers/avatar_llm_state_provider.py
```diff
@@ -91,12 +98,21 @@ def _restore_happy(self) -> None:
 
         Sets the avatar to "Happy" state after processing completion.
         """
-        if self.avatar_provider and self.avatar_provider.running:
+        if self.avatar_provider:
+            if not self.avatar_provider.running:
+                try:
+                    self.avatar_provider.start()
```
openminddev: If we use the original method, then we don't need the `start`

PR #762, opened 2025-11-25
Include basic lost and found functionalities

ShaohongZ:
## Overview
Include basic functionalities for lost and found

## Changes
1. A background process to continuously store and update sightings of objects of interest in a database;
2. A simple action to find the last sighting of an object of interest.


openminddev:
We store the data in sql, so it is a very good way for us to provide this service as the cloud service.

openminddev:
We should not run the YOLO detector in the threading, it can block others. We need a separate instance that runs the YOLO inference that other programs can interact with.

PR #770, opened 2025-11-28
Feat: multi-sensor input plugins 

ScapeXYZ:
# Documentation
**What my setup adds**
This update expands OM1 with five new real-time input plugins:

- Temperature
- Humidity
- Light intensity
- Air quality (CO₂ ppm)
- Sound level
Each plugin runs independently in its own thread and exposes a simple `.get_data()` method.
The goal is to make sensor data instantly accessible without blocking or waiting, fully compatible with future OM1 features.

## Step to run 
`(.venv) python main.py`
This will display a continuously refreshing table showing the current environment snapshot.

## Step to run the test suite
`(.venv) python -m unittest tests/test_all_sensors.py`

## Files added

`src/plugins/sensors/`
 ` src/plugins/sensors/temperature_plugin.py`
  ` src/plugins/sensors/humidity_plugin.py`
 ` src/plugins/sensors/light_plugin.py`
  ` src/plugins/sensors/air_quality_plugin.py`
 ` src/plugins/sensors/ sound_sensor_plugin.py`
  `tests/ test_all_sensors.py`
  ` main.py`

-  All sensors are mock-driven and  no hardware is required to test or demo.
- Each plugin updates continuously, so values change every loop instead of staying static.
- Designed as drop-in modules that won’t break the existing OM1 architecture.

### DIFF
--- a/main.py
+++ b/main.py
@@ -1,0 +1,49 @@
+from src.plugins.sensors.temperature_plugin import TemperaturePlugin
+from src.plugins.sensors.humidity_plugin import HumidityPlugin
+from src.plugins.sensors.light_plugin import LightSensorPlugin
+from src.plugins.sensors.air_quality_plugin import AirQualityPlugin
+from src.plugins.sensors.sound_sensor_plugin import SoundSensorPlugin
+import time
+from os import system
+
+temp = TemperaturePlugin()
+hum = HumidityPlugin()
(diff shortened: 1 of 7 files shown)


openminddev:
@ScapeXYZ Thanks for your contribution, but your code structure didn't follow the OM1 structure.

PR #797, opened 2025-12-04
fix: harden avatar provider zenoh lifecycle and payload safety

Wanbogang:
### Background

The `AvatarProvider` relies on Zenoh for real-time communication. Several safety and lifecycle gaps were identified that could impact runtime stability:

* Payload deserialization was not strictly validated
* Commands could be published when the Zenoh session was no longer valid
* Cleanup logic in `stop()` did not fully ensure safe resource release
* Logging was too generic for effective debugging in real-time systems

---

### What’s Improved

* ✅ Added strict payload validation before deserialization
* ✅ Ensured session validity before publishing commands
* ✅ Improved cleanup of publisher, subscriber, and session during `stop()`
* ✅ More specific and structured logging per critical event
* ✅ Code formatted with `black` & `isort`

---

### Why This Matters

* Prevents crashes caused by corrupted Zenoh payloads
* Improves reliability of real-time avatar communication
* Makes debugging production issues significantly easier
* No public API changes (non-breaking)

---

### Files Changed

* `src/providers/avatar_provider.py`

---

### Impact

* Improved internal reliability
* Safer real-time behavior
* Low integration risk

---

✅ This PR focuses strictly on **internal safety, robustness, and lifecycle correctness** without changing external behavior.

### DIFF
--- a/src/providers/avatar_provider.py
+++ b/src/providers/avatar_provider.py
@@ -18,17 +18,11 @@ from .singleton import singleton
 class AvatarProvider:
     """
     Singleton provider for Avatar communication via Zenoh.
-
     """
 
     def __init__(self):
-        """
-        Initialize the AvatarProvider.
-        """
         self.session = None
-        # Face Publisher
         self.avatar_publisher = None
-        # Health Check Publisher and Subscriber
         self.avatar_healthcheck_publisher = None
         self.avatar_subscriber = None
         self.running = False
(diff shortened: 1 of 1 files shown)


src/providers/avatar_provider.py
```diff
@@ -133,7 +140,28 @@ def stop(self):
 
         self.running = False
 
-        if self.session:
-            self.session.close()
+        try:
+            if self.avatar_subscriber:
+                self.avatar_subscriber.undeclare()
```
openminddev, line 145: This method doesn't exist.

Wanbogang: Thanks for the review! You’re right, undeclare() doesn’t exist on the subscriber object. I’ll update it to close() and resend the PR. Appreciate the quick feedback.

PR #823, opened 2025-12-09
support dynamic transition rules extracting

YuchengZhou821:
This pull request enhances the schema generation process in `scripts/generate_schema.py` by adding support for extracting and including transition rule schemas. The main changes introduce a new method for scanning transition rules from an external schema file and update the generated schema to include this information.

**Transition rule schema extraction:**

* Added the `scan_transition_rules` method to extract the `transition_rules` schema from `multi_mode_schema.json`, including required fields, types, and properties.

**Schema generation updates:**

* Updated the `generate` method to call `scan_transition_rules` and include the extracted transition rules in the output schema dictionary.

### DIFF
--- a/scripts/generate_schema.py
+++ b/scripts/generate_schema.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python3
 """Generate OM1 configuration schema from codebase."""
 
 import ast


scripts/generate_schema.py
```diff
@@ -251,6 +253,46 @@ def _scan_hook_functions(self) -> List[Dict[str, Any]]:
...
+    def scan_transition_rules(self) -> Dict[str, Any]:
+        """Extract transition_rules schema from multi_mode_schema.json.
+
+        Returns
+        -------
+        Dict[str, Any]
+            Transition rules schema with types and properties.
+        """
+        import json
+
```
openminddev: Please move it to the top.

openminddev (approved):

PR #876, opened 2025-12-31
fix: log naming consistency for SimplePathsProvider

ScapeXYZ:
The logger uses SimplePathProvider (singular), but the class is SimplePathsProvider (plural). This causes logs to be missed when filtering by class name. Fixed the string to ensure all logs are captured during debugging.

Old: logging.info(`"Zenoh is open for SimplePathProvider"`)
New: logging.info(`"Zenoh is open for SimplePathsProvider"`)


Wanbogang:
Hi @ScapeXYZ This issue is valid and the fix is ​​100% correct! 👍

openminddev:
@ScapeXYZ Thank you. Merged.

PR #905, opened 2026-01-02
Fix launch command and add usage instructions

volkan1627:
Updated command to launch OM1 with correct Python syntax and added additional help information.


volkan1627:
if not work run try to new run code

volkan1627:
Clarify preferred method for API key configuration
Added note about using .env file for sensitive info.

PR #907, opened 2026-01-02
feat: Add missing conversation example code

DaniuXie:
I noticed the conversation example referenced in the documentation was missing from the repository structure. I have reconstructed the standalone python script based on the docs and verified it works locally. This makes it easier for new developers to test the SDK.


Prachi1615:
Hi @DaniuXie 
https://github.com/OpenMind/OM1/blob/main/config/conversation.json5
refer here

PR #908, opened 2026-01-02
feat(spot): add GoogleASRInput to enable voice interaction in simulation

OxHanta:
### What this does
This PR enables voice interaction for the Spot simulator by adding `GoogleASRInput` to `config/spot.json5`.

This allows users running OM1 in simulation-only environments (including WSL) to speak to Spot and receive spoken responses without physical robot hardware.

### Why this is needed
- The default Spot config does not include an audio input source
- Simulation users cannot interact via voice out of the box
- Adding GoogleASRInput unlocks a key interaction path already supported by the runtime

### What changed
- Added `GoogleASRInput` to Spot agent inputs
- No breaking changes
- No hardware dependency introduced

### Tested
- Tested locally in simulation
- Confirmed microphone input → LLM → TTS response loop works


openminddev:
Hi @OxHanta `spot.json5` is an example. Thanks.

OxHanta:
Thanks for the clarification @openminddev 
I’ll follow up with a docs/example-only PR to demonstrate enabling GoogleASRInput for simulation use cases... unless someone has done that already

PR #1009, opened 2026-01-05
fix: Support multi-parameter actions in function calling

0xbyt4:
## Overview
This PR fixes a bug where function call arguments from the LLM were being lost during action processing. Only the `action` parameter was being passed to connectors, while other parameters (like `to_address`, `amount`, `chain` for wallet actions) were discarded.

### Core Fix
- **`src/llm/output_model.py`**: Added `args` field to `Action` class to store all function call arguments
- **`src/llm/function_schemas.py`**: Updated `convert_function_calls_to_actions()` to populate the `args` field with all parameters
- **`src/actions/orchestrator.py`**: Updated `_promise_action()` to use `args` when available, falling back to legacy behavior when `args` is `None`

### Test Updates
- Added 7 new tests in `tests/llm/test_function_schemas.py` for multi-parameter action conversion
- Added 4 new tests in `tests/actions/test_orchestrator.py` for args passthrough functionality
- Updated 5 LLM plugin tests to expect the new `args` field in Action objects

## Testing

**Unit Tests:**
- All 60 related tests pass
- Pre-commit hooks (ruff, black, isort, typos, etc.) all pass

**Test Commands:**
```bash
uv run pytest tests/llm/test_function_schemas.py tests/actions/test_orchestrator.py -v
uv run pre-commit run --all-files
```

### Backward Compatibility
This change is **fully backward compatible**:
- When `args` is `None` or empty, the orchestrator falls back to the original behavior (`{"action": action.value}`)
- Existing single-parameter actions (move, speak, etc.) continue to work unchanged

### New Capability
Actions with multiple parameters can now receive all arguments from LLM function calls:

**Before (Bug):**
```
LLM sends: {"action": "send", "to_address": "0x123", "amount": "0.5"}
Connector receives: {"action": "send"}  ❌ Parameters lost!
```

**After (Fixed):**
```
LLM sends: {"action": "send", "to_address": "0x123", "amount": "0.5"}
Connector receives: {"action": "send", "to_address": "0x123", "amount": "0.5"}  ✅
```

## Additional Information
This fix enables multi-parameter actions like wallet operations, API calls, or any action requiring more than a single `action` parameter to work correctly with OM1's function calling system.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,11 +1,24 @@
 ## Overview
 [Provide a brief overview of the changes in this pull request.]
+(If applicable, linked issue: [ ])
+
+## Type of change
+- [ ] Bug fix
+- [ ] New feature
+- [ ] Documentation improvement
+- [ ] Bounty issue submission
+- [ ] Other:
 
 ## Changes
 [Detail the changes you have made in this pull request. Include any new features, bug fixes, or improvements.]
 
-## Testing
-[Describe the testing you have done to verify your changes. Include any relevant details about the test environment and the results of your tests.]
+## Checklist
+
+- [ ] Code complies with style guidelines
+- [ ] Self-review completed
+- [ ] Documentation updated (if applicable)
+- [ ] Local testing completed
+- [ ] Tests added/updated (if applicable)
 
 ## Impact
 [Explain the impact of your changes. Include any potential risks or side effects that reviewers should be aware of.]
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,6 +4,8 @@ We welcome contributions from the community!  OM1 is an open-source project, and
 
 Before contributing, please take a moment to read through the following guidelines. This helps streamline the process and ensures everyone is on the same page.
 
+**PRs must clearly state the problem being solved. Changes without a clear problem statement may be closed without review.**
+
 **Ways to Contribute:**
 
 *   **Report Bugs:** If you find a bug, please [open an issue](https://github.com/OpenMind/OM1/issues) on GitHub. Be sure to include:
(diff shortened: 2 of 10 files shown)


openminddev:
Hi @0xbyt4, we support multiple args. You need to tell the LLM the examples about the variables. https://github.com/OpenMind/OM1/blob/7b3301c5011e03aec7c26fcbac82ab9245304d34/src/actions/remember_location/interface.py#L31

0xbyt4:
Hi @openminddev, thank you for the feedback! I appreciate you taking the time to review.

I understand that interfaces can define multiple parameters (like `remember_location` with `description`). However, I believe the bug I identified is in the **code flow**, not the interface definition.

## The Issue

Let me trace through the actual code path:

### Step 1: LLM generates function call
```json
{"name": "wallet", "arguments": "{\"action\":\"send\",\"to_address\":\"0x123\",\"amount\":\"0.5\"}"}
```

### Step 2: `convert_function_calls_to_actions()` in `function_schemas.py`
```python
# Line 156-170
action_value = args.get("action", "")  # Gets "send"
action = Action(type=function_name, value=action_value)  # Only type and value!
```
**Result:** `Action(type="wallet", value="send")` — `to_address` and `amount` are discarded here.

### Step 3: `_promise_action()` in `orchestrator.py`
```python
# Line ~220
input_interface = T.get_type_hints(agent_action.interface)["input"](
    **{"action": action.value}  # Only passes action.value!
)
```
**Result:** `WalletInput(action="send")` — only the `action` field is passed.

## Why `remember_location` Works

The `remember_location` example works because `description` has a **default value**:
```python
description: Optional[str] = ""  # Has default, so it's OK if not passed
```

But for an action with **required** parameters (no defaults):
```python
@dataclass
class WalletInput:
    action: str
    to_address: str  # Required, no default!
    amount: str      # Required, no default!
```
This would either:
- Raise `TypeError: missing required argument`
- Or the connector receives incomplete data

## My Fix

This PR addressed both issues:
1. Added `args` field to `Action` model to store all parameters
2. Updated `convert_function_calls_to_actions()` to populate `args`
3. Updated `_promise_action()` to use `args` when available (with backward compatibility)

Would you be open to reconsidering, or am I missing something in the architecture?

PR #1011, opened 2026-01-05
Add JSON5 config linting with GitHub Actions

ScapeXYZ:
Adds a Python-based JSON5 linter and CI workflow that validates all `config/*.json5`
`files and fails PRs if any config is invalid`.


openminddev:
Hi @ScapeXYZ, please refer to #1039 as an example when you create a PR.

Issue #985, opened 2026-01-05
Save the conversation history to disk and reload it if it is exists


openminddev:
Save conversation history to disk; restore on crash/restart.

Wanbogang:
@openminddev 

This issue highlights the need for persistence, but currently there is no canonical in-memory conversation history to persist. At the moment, prompts and LLM outputs are generated and consumed per tick without a centralized conversation layer. I’m working on a small, non-invasive PR that introduces an in-memory conversation history as a foundation, which would make disk persistence and crash/restart recovery (as described here) straightforward and well-scoped.

I’d appreciate your thoughts on this approach — does introducing this foundational layer first align with how you envision solving this issue?

goodluxiao2:
I'd like to work on this. I can implement a local JSON/SQLite based history persistence layer for OM1.

goodluxiao2:
I've drafted a persistence layer using JSONL for chat history. This will ensure conversations are saved locally and reloaded automatically. Ready to submit a PR once assigned.

lau90eth:
Hi @openminddev   
I’ve implemented persistent conversation history with automatic reload on startup.

The solution introduces a centralized in-memory history layer, persists it to disk,
and reloads it on runtime restart, while preventing prompt explosion.

PR #1030 
Happy to make any adjustments if needed.

openminddev:
We are looking for a simple JSON storage solution. The current chat history is a json dictionary.

Wanbogang:
Hi @openminddev 

I've read your comment about wanting a "simple JSON storage solution" and realized my initial implementation #1014  might be too complex.

I've created a new alternative, #1049 

Both PRs are ready for review. The new one better aligns with your request for simplicity, but I'm happy to proceed with either approach based on your preference.

Bara69chn:
Hi team, I’d like to work on this bounty (#985). I noticed there are already PRs linked on the right. Is there any remaining work needed (e.g., tests, edge cases, docs), or should I pick a different bounty? If it’s still open, I can implement a simple JSON-based save/load for the conversation history as requested

lau90eth:
Hi @openminddev 

I’ve aligned the implementation to a *very simple JSON-based approach*, as requested.

What this PR #1055  does:
- Keeps conversation history as a plain JSON-compatible structure (`list[dict]`)
- No additional dependencies or abstractions
- Messages are always `{role: str, content: str}` (OpenAI-safe)
- History length is bounded by `history_length`
[...]

PR #1030, opened 2026-01-06
feat(history): persist conversation history and reload on startup

lau90eth:
Implements persistent conversation history for OM1.

Conversation history is saved to disk and automatically reloaded on restart,
allowing the agent to resume context after crashes or restarts.

This addresses bounty issue #985.

### DIFF
--- a/src/llm/plugins/openai_llm.py
+++ b/src/llm/plugins/openai_llm.py
@@ -1,4 +1,5 @@
 import logging
+import os
 import time
 import typing as T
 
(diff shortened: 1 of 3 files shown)


lau90eth:
Hi @openminddev 
I implemented conversation history persistence (save to disk + reload on restart)
and opened a PR earlier, which was closed.

I can re-submit a cleaner/minimal PR focusing only on the history-related files.
Let me know if you'd like me to proceed that way

PR #1093, opened 2026-01-07
add missing stop() methods to Gemini and OpenAI plugins

Savage890:
I noticed that the VLMGemini and VLMOpenAI inputs weren't cleaning up their providers when stopped. This could leave background threads or camera streams running even after the input is supposed to be disabled.
I've added stop() methods to both classes to ensure they properly delegate the cleanup to their underlying providers, similar to how other plugins like GoogleASRInput handle it.

### DIFF
--- a/config/schema/single_mode_schema.json
+++ b/config/schema/single_mode_schema.json
@@ -15,7 +15,11 @@
     ],
     "properties": {
         "version": {"type": "string"},
-        "hertz": {"type": "number"},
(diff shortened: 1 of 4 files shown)


openminddev:
Hi @Savage890 please fix the lint and follow the code style.

Savage890:
ok wait

Savage890:
Done

PR #1097, opened 2026-01-07
Doc tweaks

Prachi1615:
## Overview
Updated development guide

### DIFF
--- a/docs/developer_cookbook/introduction.mdx
+++ b/docs/developer_cookbook/introduction.mdx
@@ -19,14 +19,10 @@ Then dive into any recipe that interests you!
 
 ## Development workflow
 
-### Linting and Testing
+### Linting and Testing (Mandatory)
 
-To check/format/lint your code, run:
-```bash
-uv run ruff check . --fix && uv run black . && uv run isort .
-```
+To check/format/lint your code before committing, install [pre-commit](https://pre-commit.com/) and execute `pre-commit install`. This ensures that pre-commit checks run before each commit.
 
-To automatically run these checks before committing, install [pre-commit](https://pre-commit.com/) and execute `pre-commit install`. This ensures that pre-commit checks run before each commit.
 Additionally, you can manually trigger all checks by running
 
 ```bash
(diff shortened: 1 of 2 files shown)


docs/developing/1_get-started.mdx
```diff
@@ -220,10 +218,10 @@ INFO:root:OpenAI LLM output: commands=[Command(type='move', value='wag tail'), C
 
 There are more pre-configured agents in the `/config` folder. They can be run with the following command:
 
-For example, to run the `conversation` agent:
```
openminddev, line 223: Should we keep the conversation agent as an example? It is more interesting than the cubly.

docs/developer_cookbook/introduction.mdx
```diff
@@ -10,3 +10,40 @@ Here's what you can do with OM1
...
+Then dive into any recipe that interests you!
+
+## Development workflow
+
+### Linting and Testing
+
+To check/format/lint your code, run:
+```bash
+uv run ruff check . --fix && uv run black . && uv run isort .
+```
```
openminddev: I suggest requiring users to install pre-commit for linting and initial checks, as we may integrate additional tools in the future.

docs/developing/1_get-started.mdx
Prachi1615: conversation is already mentioned a an example under switching agents. wouldn't it be better to refer another one?
alternatively can mention any other too

docs/developer_cookbook/introduction.mdx
Prachi1615: should we remove this section itself?
pre-commit installation is below this

docs/developing/1_get-started.mdx
openminddev: Ok! Got it.

docs/developer_cookbook/introduction.mdx
openminddev: Yes, please make the `pre-commit` installation as a requirement if you want to do the initial checks.

openminddev (approved):

PR #1099, opened 2026-01-07
fix: fix CycloneDDS network interface config for Linux compatibility

shashiyy:
Ran into DDS communication failures on my Linux dev box. The cyclonedds.xml was hardcoded to use `en0` which is a macOS interface name. On Linux we have eth0, ens0, eno1, etc.

Updated the config to support common interface names across different platforms - macOS, Linux, Docker, and VMs. Also threw in some performance settings for handling larger robot data like point clouds.

Tested on Ubuntu and it connects fine now.

### DIFF
--- a/cyclonedds/cyclonedds.xml
+++ b/cyclonedds/cyclonedds.xml
@@ -1,12 +1,92 @@
-<CycloneDDS>
-  <Domain>
+<?xml version="1.0" encoding="UTF-8" ?>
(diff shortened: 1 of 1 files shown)


openminddev:
Hi @shashiyy thanks for your PR. We provide the minimum interface for users. You can modify the network interface to make it work. We don't have any plan to support all network interfaces.

PR #1101, opened 2026-01-08
fix(providers): prevent empty sensor prompts from triggering LLM

### LINKED ISSUE #1025: Spot repeatedly sends empty sensor prompts causing rapid LLM requests and 429 rate limiting
### Description
When running Spot with the default configuration (`uv run src/run.py spot`),
the agent repeatedly sends prompts with empty sensor content:

"Spot sensed the following: "

This results in:
- Rapid repeated requests to the LLM endpoint
- HTTP 429 Too Many Requests responses
- Unnecessary API usage and retry loops

### Evidence
Logs show repeated messages with empty payload being sent to the LLM,
followed by 429 responses and immediate retries.

### Expected behavior
- Spot should skip LLM calls when sensor output is empty
- There should be a minimum cooldown or debounce between requests
- Empty sensor data should not trigger LLM calls

### Notes
The Spot behavior appears to be runtime-driven; this repository exposes
configuration only, and there is currently no configurable guard against
empty sensor input.

Savage890:
Fix: Spot sometimes spams empty prompts causing 429s

Hey! I looked into Issue #1025 where Spot keeps sending "Spot sensed the following: " with nothing else, which drives up API usage and hits rate limits.

I found that `LLMHistoryManager` was constructing the prompt unconditionally. I added a check to ensure we actually have new sensor data for the current tick before firing off a request to the LLM. If there's no new data, it just skips the call.

This should stop the empty request loop and fix the 429 errors!

Fixes #1025

### DIFF
--- a/.github/workflows/nightly-integration-tests.yml
+++ b/.github/workflows/nightly-integration-tests.yml
@@ -1,4 +1,4 @@
-name: Run Integration Tests
+name: Run Nightly Integration Tests
 
 env:
   OM1_API_KEY: ${{ secrets.OM1_API_KEY }}
@@ -15,7 +15,7 @@ on:
       - '**'
 
 jobs:
-  integration-test:
+  nightly-integration-test:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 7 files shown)


openminddev:
Hi @Savage890, you can change the hertz to increase the ticker duration.

Issue #984, opened 2026-01-05
Improve hot-reload configs


openminddev:
Now, we don't support to update a certain fields without restarting the whole system. It makes sense to load certain config fields (e.g., system_prompt_base) without restarting; watch for file changes.

lau90eth:
Hi @openminddev 

This PR implements a selective and safe hot-reload mechanism for runtime
configuration files, allowing specific fields (e.g. `system_prompt_base`)
to be updated at runtime without requiring a full system restart.

The solution introduces:
- A watchdog-based config file watcher
- An explicit whitelist of hot-reloadable fields
- Backward-compatible integration with the existing runtime lifecycle

The scope of the PR is intentionally limited to hot-reload improvements only.

Closes #984  
Working on this bounty with PR #1031

openminddev:
We are looking for a solution where https://github.com/OpenMind/OM1/blob/3ca608d419a44f6e71db844da7e1909cbfe74368/src/runtime/single_mode/cortex.py#L221 can reload based on the changed fields.

lau90eth:
Yes, that makes sense  
I can add a runtime-level hook in `CortexRuntime` so that reload decisions
are made based on the set of changed fields (partial vs full reload).
I’ll follow up with a revised PR.

Bara69chn:
Hi, Dev. I’d like to help on this bounty #984. I see PRs #1020/#1031 working on selective reload. I can help by testing the hot-reload behavior (which fields trigger partial vs full reload), adding edge-case handling (invalid JSON5, rapid edits), and adding unit tests/docs if needed. Please point me to the expected field whitelist or desired behavior

Wanbogang:
Hi @openminddev 
I've submitted a PR implementing the selective hot-reload feature with watchdog.

**Key features:**
- Watchdog-based monitoring (no more polling)
- Selective reload for system prompts and LLM params
- Critical fields (model, provider, api_key) marked as restart-required
- Backward compatible

PR: #1061 

Looking forward to your feedback

lau90eth:
Thanks @openminddev for the recent PRs and discussion around selective hot-reload.

Just to clarify alignment with the original maintainer feedback on this issue:
the key requirement was that reload decisions live explicitly in
`src/runtime/single_mode/cortex.py`, and that CortexRuntime itself decides
partial vs full reload based on the set of changed fields.

I’ve submitted a revised PR ( #1064 ) that focuses specifically on that point:
[...]

PR #1108, opened 2026-01-08
fix: prevent re-entrant hot-reload during config reload 

Bara69chn:
## Overview

- Add a small guard to prevent re-entrant config reloads while a reload is already in progress. 
- This improves safety for the hot-reload loop in `src/runtime/single_mode/cortex.py`: if the config file changes again during an ongoing reload (e.g., rapid edits), we skip triggering another reload until the current one finishes.

Related to: #984

## Type of change
- [x] Bug fix 
- [x] Bounty issue submission

## Changes
- Skip calling `_reload_config()` in `_check_config_changes()` when `_is_reloading` is `True`.

## Reproduction steps
1. Run OM1 in single mode with hot-reload enabled and a small `check_interval`.
2. While a reload is in progress, edit/save the config file multiple times quickly (rapid edits).
3. Without the guard, reload may be triggered back-to-back / re-entrantly.
4. With this change, additional triggers are ignored until the current reload completes.

## Test plan
- `uv run python -m py_compile src/runtime/single_mode/cortex.py`
- `uv run pre-commit run --all-files`
- `uv run pytest -q --ignore=tests/inputs/base/test_inputs_plugins.py`

## Impact
Low risk. This only prevents triggering a new reload while a reload is already running. Normal reload behavior remains unchanged.


openminddev:
@Bara69chn Please add the unit test to verify that the issue has been resolved.

Bara69chn:
@openminddev  Added a unit test (`tests/runtime/test_hot_reload_guard.py`) to verify that `_check_config_changes()` skips calling `_reload_config()` when `_is_reloading` is `True`. Pre-commit and the test pass locally.

Bara69chn:
I fixed ruff/pre-commit import order in tests/runtime/test_hot_reload_guard.py and force-pushed. CI should be green now

0xbyt4:
Hi @Bara69chn  _check_config_changes awaits _reload_config sequentially in the same async task, so the while loop cannot advance while a reload is running ,this guard can never trigger ?  and This PR also doesn't address #984

PR #1123, opened 2026-01-08
fix(schema): enforce positive history_length in both config schemas

Wanbogang:
## Problem
Both single-mode and multi-mode schemas allowed zero or negative `history_length` values, which would result in invalid conversation history configurations.

## Solution
Added `minimum: 0` constraint to the `history_length` field in both schema files, ensuring at least one history entry is maintained.

## Changes
- Added `minimum: 0` to history_length in single-mode schema
- Added `minimum: 0` to history_length in multi-mode schema (2 locations)
- Added descriptive comment for clarity

## Impact
Prevents configuration errors where history_length is set to 0 or negative values, which were found in some multi-mode configs.

### DIFF
--- a/src/backgrounds/plugins/odom.py
+++ b/src/backgrounds/plugins/odom.py
@@ -30,10 +30,33 @@ class OdomConfig(BackgroundConfig):
 
 class Odom(Background[OdomConfig]):
     """
-    Reads odometry data from Odom provider.
+    Background task for reading odometry data from Odom provider.
+
+    This background task initializes and manages an OdomProvider instance
+    that reads robot odometry data. The provider can connect via Zenoh
+    for distributed robot systems or via Unitree Ethernet for direct
+    hardware communication.
+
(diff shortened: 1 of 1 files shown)


openminddev:
Hi @Wanbogang, it can be zero, so it must be larger or equal to zero.

Wanbogang:
> Hi @Wanbogang, it can be zero, so it must be larger or equal to zero.

Thanks for the feedback! You're right - after reviewing the configuration, I see that `history_length: 0` is intentionally used for stateless mode.

While `minimum: 0` will still prevent negative values, I agree that its practical value is minimal since negative history lengths are unlikely in real-world configurations.

Have I changed this to 0 , Or should I close this PR?

openminddev:
Hi @Wanbogang , I don't think the negative value is possible. I will close this PR for now.

Wanbogang:
Hi @openminddev .Yes, you are right, sir. Thank you.

PR #1151, opened 2026-01-08
perf: clean up mock object setup in tests

MonkeyMarcel:
found a few test files using __dict__.update() for setting mock attributes and it seemed like a cleaner approach than doing it one by one. updated similar patterns in runtime and simulator tests to match.

touched 4 test files


openminddev:
Hi @MonkeyMarcel, thanks for your PR, but you overwrote the lint. Thanks.

PR #1193, opened 2026-01-09
chore(schema): deduplicate cortex_llm definition using shared ref

MuhamadRifansyah:
# Overview
This PR refactors the `multi_mode_schema.json` configuration schema to
eliminate code duplication by introducing a shared definition for the
`cortex_llm` object. The shared definition is reused in both the global
configuration and individual mode definitions.

# Changes
- Introduced a shared schema definition for `cortex_llm` under
  `definitions.cortex_llm`.
- Replaced the inline `cortex_llm` schema definitions in:
  - the root-level `properties`
  - the `modes` pattern properties
  with `$ref` references to the shared definition.

# Impact
## Maintenance
Reduces the risk of schema drift by ensuring future updates to the LLM
configuration structure need to be applied in only one place.

## Consistency
Ensures that global and mode-specific LLM configurations are validated
against the exact same schema rules.

# Testing
- [x] Verified that the schema remains valid JSON.
- [x] Confirmed that the structure of the `cortex_llm` object is preserved
      in the new shared definition.

# Additional Information
Addresses the **Duplicate Schema Definitions** finding.

### DIFF
--- a/.github/workflows/nightly-integration-tests.yml
+++ b/.github/workflows/nightly-integration-tests.yml
@@ -38,7 +38,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: /tmp/cyclonedds/cyclonedds/install
-          key: ${{ runner.os }}-cyclonedds-0.10.x-${{ hashFiles('.github/workflows/integration-test.yml') }}
+          key: ${{ runner.os }}-cyclonedds-0.10.x-${{ hashFiles('.github/workflows/nightly-integration-tests.yml') }}
           restore-keys: |
             ${{ runner.os }}-cyclonedds-0.10.x-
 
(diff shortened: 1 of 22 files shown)


openminddev:
Hi @MuhamadRifansyah , json validation fails.

MuhamadRifansyah:
> Hi @MuhamadRifansyah , json validation fails.

Hi @openminddev, thanks for the heads-up.

I’ve pulled the latest changes from main and fixed the Draft-07 structural nesting issue in multi_mode_schema.json that was causing the JSON schema validation failure.
The fix is limited to correcting object/patternProperties closure and $ref placement only, no semantic or validation rule changes.
CI has been re-triggered and should now pass for schema validation.
Thanks!

openminddev (approved):

PR #1272, opened 2026-01-10
fix: resolve critical race conditions and resource safety issues

0xbyt4:
## Summary
- **Bug 1:** Race condition in mode transition - Extended lock scope to cover entire try/except/finally block
- **Bug 2:** Zenoh publisher undefined after init failure - Fixed incorrect variable name and added null checks
- **Bug 3:** LLM history manager thread-safety - Added threading.Lock for message list modifications

## Changes
- `src/runtime/multi_mode/manager.py`: Fixed race condition by extending async lock scope, corrected `self.pub` to `self._zenoh_mode_status_response_pub`, added null checks at lines 755 and 825
- `src/providers/llm_history_manager.py`: Added `threading.Lock` for thread-safe list operations, replaced unsafe `pop(0) if messages else None` with proper length checks

## Test plan
- [x] All existing tests pass (648 passed, 1 skipped due to missing ubtech module)
- [x] No breaking changes - all modifications are backwards compatible
- [x] Follows existing code patterns (threading.Lock used in 14+ other files)

### DIFF
--- a/src/providers/llm_history_manager.py
+++ b/src/providers/llm_history_manager.py
@@ -1,6 +1,7 @@
 import asyncio
 import functools
 import logging
+import threading
 from dataclasses import dataclass
 from typing import Any, Awaitable, Callable, List, Optional, TypeVar, Union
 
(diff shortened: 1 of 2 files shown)


openminddev:
Hi @0xbyt4 thanks for your PR. Please split your PR into three PRs for different problems and add the unit tests to cover the edge case.

PR #1279, opened 2026-01-11
fix: define placeholder dds_ module in unitree_g1_basic.py

Wanbogang:
Fixes a `NameError: name 'dds_' is not defined` that occurred when importing `unitree_g1_basic.py` if the optional Unitree SDK dependency was not installed. This happened because the type hints `dds_.BmsState_` and `dds_.LowState_ ` were evaluated during class definition, but `dds_` was not defined in the `except ImportError` block.

This change defines `dds_` as a placeholder module containing placeholder classes (`PlaceholderBmsState_, PlaceholderLowState_`) within the `except` block, ensuring the type hints are valid and preventing the import failure.

### DIFF
--- a/src/inputs/plugins/unitree_g1_basic.py
+++ b/src/inputs/plugins/unitree_g1_basic.py
@@ -17,6 +17,49 @@ except ImportError:
         "Unitree SDK not found. Please install the Unitree SDK to use this plugin."
     )
 
+    # --- Fix: Define a placeholder dds_ module ---
+    # Create a fake 'dds_' module containing placeholder classes.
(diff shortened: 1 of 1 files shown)


0xbyt4:
https://github.com/OpenMind/OM1/pull/1217 already here and closed ,   actually it is not a bug  or no need to fix .  if you install extra needed  it passes

Wanbogang:
Hi @0xbyt4   Indeed, the problem faced is the same as the PR you created, but our approach is different. So in my opinion, let @openminddev  decide whether this needs to be closed or not.

PR #1289, opened 2026-01-11
fix: add timeout and error handling to arm connector

windlgrass:
## Summary

Adds timeout and error handling to arm connector to prevent system crashes and indefinite hangs.

## Problem

The current implementation has two critical issues:

### 1. No Error Handling
When robot hardware fails or connection is lost, `ExecuteAction()` raises an exception that crashes the entire system instead of being handled gracefully.

**Example:**
```
Robot executing "high five" → Cable unplugged → ConnectionError → 
System crashes → Requires manual restart
```

### 2. No Timeout
If the robot gets stuck (motor jam, mechanical issue), the code waits indefinitely, freezing the application.

**Example:**
```
Robot starts "clap" → Motor stuck → Code hangs forever → 
System frozen, cannot execute other commands
```

### Added Error Handling
```python
try:
    await asyncio.wait_for(...)
except asyncio.TimeoutError:
    logging.error("Action timed out", exc_info=True)
except ConnectionError as e:
    logging.error(f"Connection lost: {e}", exc_info=True)
except Exception as e:
    logging.error(f"Action failed: {e}", exc_info=True)
```

### Added 10-Second Timeout
```python
await asyncio.wait_for(
    loop.run_in_executor(None, self.client.ExecuteAction, action_id),
    timeout=10.0
)
```

## Impact

**Before:**
- ❌ System crashes on hardware failure
- ❌ Code hangs indefinitely if robot stuck
- ❌ Requires manual restart and causes downtime

**After:**
- ✅ Graceful error handling - system continues on failure
- ✅ Timeout prevents hangs - system remains responsive
- ✅ Better logging with full tracebacks for debugging

## Testing

- Normal actions (3-5s) complete successfully with no performance impact
- Timeout triggers after 10s if robot stuck, system continues operating
- Connection errors are logged gracefully without crashing

This makes the connector production-ready and prevents critical failures.


openminddev:
Hi @windlgrass this problem comes from Unitree SDK. The right way to avoid this is to use `CallNoReply` in the sdk.

windlgrass:
Thank you for the feedback! I understood CallNoReply is the SDK's recommended approach. Will keep this in mind for future contributions. 🙏

PR #1291, opened 2026-01-11
fix: add thread safety to LLM history manager with asyncio.Lock

Wanbogang:
This PR adds `asyncio.Lock` protection around all accesses to the LLM history buffer to prevent race conditions during concurrent async operations (e.g., summarization and new message ingestion).

- Uses `async with self._history_lock` for all reads/writes to `self.history`
- Adds safe helper methods: `add_message`, `clear_history`, `get_messages`
- Ensures callback updates from background tasks are scheduled back into the main event loop via `asyncio.run_coroutine_threadsafe`

No functional changes — only internal robustness improvement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,36 +147,36 @@ command: ["unitree_go2_autonomy_advance"]
 
 ```bash
 cd OM1
-docker compose up om1 -d --no-build
+docker-compose up om1 -d --no-build
 ```
 
 - For unitree_sdk
 ```bash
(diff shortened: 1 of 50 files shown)


0xbyt4:
https://github.com/OpenMind/OM1/pull/1277 already here

Wanbogang:
> #1277 already here

Hi @0xbyt4 Your understanding and mine are different sir in terms of approach.Thanks

PR #1338, opened 2026-01-12
docs(simple_paths): clarify parameters and init

haidao1919:
Fixed docstring formatting.

Updated comments for clarity.


Wanbogang:
@haidao1919 
Thank you for your contribution. The documentation is too verbose and doesn't provide significant value.
There's a lot of redundant information that's already clear from the code. The original is sufficient.

PR #1362, opened 2026-01-12
fix: use asyncio.to_thread for non-blocking HTTP in fabric_closest_peer

0xbyt4:
## Summary
- Fix blocking `requests.post()` call in async `_poll()` method
- Wrap HTTP call with `asyncio.to_thread()` to prevent event loop blocking

## Problem
`fabric_closest_peer.py` uses `requests.post()` directly in an async function, which blocks the event loop during HTTP requests. This can cause delays in other async tasks.

Detected by: `ruff check --select=ASYNC` (ASYNC210)

## Solution
Replace:
```python
resp = requests.post(...)
```

With:
```python
resp = await asyncio.to_thread(requests.post, ...)
```

## Test Plan
- [x] Add unit tests for async HTTP behavior
- [x] All tests pass
- [x] Pre-commit hooks pass
- [x] `ruff check --select=ASYNC` passes (no ASYNC210 error)

### DIFF
--- a/src/inputs/plugins/fabric_closest_peer.py
+++ b/src/inputs/plugins/fabric_closest_peer.py
@@ -4,7 +4,7 @@ import time
 from queue import Queue
 from typing import List, Optional
 
-import requests
+import aiohttp
 from pydantic import Field
 
 from inputs.base import SensorConfig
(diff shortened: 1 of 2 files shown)


openminddev:
Hi @0xbyt4 Could you please update it using `aiohttp`?

PR #1388, opened 2026-01-13
fix(cli): surface SyntaxError during component validation

MuhamadRifansyah:
## Overview 
This PR improves the validate_config command by explicitly catching and logging SyntaxError when parsing component files. Previously, the AST parser wrapped all exceptions in a broad try...except Exception block, causing syntax errors in plugin files to be silently swallowed. This led to confusing "component not found" errors instead of identifying the actual syntax issue.

## Changes
Modified src/cli.py:
- Added a specific except SyntaxError block in _check_class_in_dir.
The new block prints a warning with the filepath and error details to stdout.
- Added tests/runtime/test_cli_validation.py:
Created a new test file to verify that SyntaxError in a plugin file triggers a warning.
Verifies that valid components in the same directory are still correctly identified despite sibling errors.

## Impac:
- Diagnostics: Users will now see a clear warning if their custom plugin files contain syntax errors, rather than a misleading "not found" error.
- UX: Slightly alters CLI output when an error is found, but ensures critical information is not hidden.
- Risk: Low. The validation logic continues after logging the error, preserving existing behavior for other files.

## Testing:
- Added tests/runtime/test_cli_validation.py.
- Verified that SyntaxError in a dummy .py file triggers the warning during execution.
- Verified that valid files are still correctly identified.

## Aditional Information
This change helps new contributors/users debugging their custom plugins by surfacing python syntax errors immediately during the config validation step.

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -610,6 +610,9 @@ def _check_class_in_dir(directory: str, class_name: str) -> bool:
                     for node in tree.body:
                         if isinstance(node, ast.ClassDef) and node.name == class_name:
                             return True
+            except SyntaxError as e:
+                print(f"\nWarning: Syntax error parsing {filepath}: {e}")
+                continue
             except Exception:
                 continue
     return False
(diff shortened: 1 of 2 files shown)


0xbyt4:
Hi @MuhamadRifansyah . Duplicate , this test already exists.

PR #1397, opened 2026-01-13
fix(actions): add type hints and validation for move_to_peer interface

longjin0593:
Hi! 👋 I noticed the `move_to_peer` interface was missing type hints and parameter validation, so I added them to improve code safety! ✨

Here's what I did:
- Added `from __future__ import annotations` for better type hint support
- Enhanced the `MoveToPeerAction` enum docstring with Attributes section (documented IDLE and NAVIGATE actions)
- Implemented `__post_init__` method in `MoveToPeerInput` to validate the `action` parameter (checks for None and ensures it's a valid enum value)
- Expanded the `MoveToPeer` class docstring with more details about peer-to-peer navigation capabilities

This should help prevent runtime errors when using the `action` parameter. Hope this helps! 🚀


Wanbogang:
@longjin0593 
Please take a look and fix the test errors

PR #1434, opened 2026-01-14
fix: add JSON parsing error handling in nav2 and slam hooks

Vt01nft:
## Summary
Adds try/except error handling for JSON parsing in success response paths.

## Bug
In `nav2_hook.py` and `slam_hook.py`, the `response.json()` calls in the `status == 200` success paths had no error handling. If the server returns status 200 but malformed/empty JSON, a `JSONDecodeError` would crash the function.

The error paths already had proper try/except handling, but the success paths did not.

## Fix
Wrap all `response.json()` calls in try/except blocks with sensible defaults, matching the existing pattern used in error paths.

## Changes
- `src/hooks/nav2_hook.py` - 2 fixes (start_nav2_hook, stop_nav2_hook)
- `src/hooks/slam_hook.py` - 3 fixes (start_slam_hook, stop_slam_hook with save + stop)

### DIFF
--- a/src/hooks/nav2_hook.py
+++ b/src/hooks/nav2_hook.py
@@ -89,11 +89,11 @@ async def stop_nav2_hook(context: Dict[str, Any]):
                     except Exception as _:
                         result = {"message": "Success"}
                     logging.info(
-                        f"Nav2 started successfully: {result.get('message', 'Success')}"
+                        f"Nav2 stopped successfully: {result.get('message', 'Success')}"
                     )
                     return {
                         "status": "success",
-                        "message": "Nav2 process initiated",
+                        "message": "Nav2 process stopped",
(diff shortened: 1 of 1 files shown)


Wanbogang:
@Vt01nft 
all error cases already have proper handling
- Error responses (non-200) already have try-except
- Network errors already caught by aiohttp.ClientError
This only adds safety for rare edge case of malformed JSON on 200 response
Thanks

Vt01nft:
> @Vt01nft 
> all error cases already have proper handling
> - Error responses (non-200) already have try-except
> - Network errors already caught by aiohttp.ClientError
> This only adds safety for rare edge case of malformed JSON on 200 response
> Thanks


Okay I get it my change was minor

PR #1469, opened 2026-01-14
fix(rpdriver): add timeout guard to prevent infinite read loop

MuhamadRifansyah:
## Overview
This PR addresses a critical reliability issue in the RPLidar driver where the read loop could block indefinitely if the hardware became unresponsive. It introduces a timeout mechanism to the serial read logic. Additionally, it adds a new suite of unit tests to verify the packet parsing logic for both normal and express scans, improving the overall test coverage of the hardware driver.

## Changes
- system_hw_test/rpdriver.py:
Updated _read_response to include a timeout check within the while loop.
The loop now raises an RPLidarException if the buffer does not fill within self.timeout seconds, preventing infinite blocking.
- tests/providers/test_rplidar_driver.py:
Created a new test file.
Added unit tests for _process_scan to verify correct parsing of quality, angle, and distance from raw bytes.
Added unit tests for ExpressPacket.from_string to verify checksum validation and data parsing for express mode packets.

## Impact
- Reliability: The driver will now fail gracefully with an exception rather than hanging the entire application if the LIDAR hardware disconnects or stops transmitting data.
- Maintainability: The complex bit-shifting logic used for parsing LIDAR packets is now covered by unit tests, reducing the risk of regression in future updates.

## Testing
- Unit Tests: Ran pytest tests/providers/test_rplidar_driver.py.
Verified that valid packets are parsed correctly.
Verified that invalid checkbits, flag mismatches, and checksum errors raise the expected exceptions.
- Manual Testing: (Optional: If you have tested this with a real RPLidar, you can mention it here, e.g., "Verified driver timeout behavior by disconnecting the device during operation.")

## Additional Information
This change focuses on the Python-side logic and does not alter the underlying serial configuration or external dependencies. The timeout implementation uses time.time() to ensure compatibility with the existing blocking loop structure.

### DIFF
--- a/system_hw_test/rpdriver.py
+++ b/system_hw_test/rpdriver.py
@@ -227,8 +227,11 @@ class RPDriver(object):
     def _read_response(self, dsize):
         """Reads response packet with length of `dsize` bytes"""
         self.logger.debug("Trying to read response: %d bytes", dsize)
+        start_time = time.time()
         while self._serial.inWaiting() < dsize:
             time.sleep(0.001)
+            if (time.time() - start_time) > self.timeout:
+                raise RPLidarException("Timeout waiting for response")
         data = self._serial.read(dsize)
         self.logger.debug("Received data: %s", _showhex(data))
         return data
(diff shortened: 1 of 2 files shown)


openminddev:
Hi @MuhamadRifansyah RPLidar doesn't work like this.

PR #1483, opened 2026-01-15
fix: correct logging name in simple_paths_processor

Vt01nft:
## Summary
Fixes copy-paste error in logging setup.

## Bug
In `simple_paths_provider.py`, the `setup_logging()` call was using `"rplidar_processor"` instead of `"simple_paths_processor"`. This causes incorrect logging labels.

## Fix
Changed logging name from `"rplidar_processor"` to `"simple_paths_processor"`.

## Changes
- `src/providers/simple_paths_provider.py` - Line 31

### DIFF
--- a/src/providers/simple_paths_provider.py
+++ b/src/providers/simple_paths_provider.py
@@ -28,7 +28,7 @@ def simple_paths_processor(
     control_queue : mp.Queue
         Queue for sending control commands.
     """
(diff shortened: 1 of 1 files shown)


Wanbogang:
@Vt01nft 
Logging name inconsistency noted, but left unchanged to keep PR scope minimal

PR #1514, opened 2026-01-16
fix(providers): implement thread-safe obstacle data access in D435Provider

xieNniu:
Added thread-safe protection for obstacle point cloud data in `D435Provider` to prevent race conditions when accessing shared state from multiple threads.

**Changes:**
- Introduced `threading.Lock` to protect read/write operations on the `obstacle` data structure
- Converted `obstacle` to private attribute `_obstacle` with a thread-safe `@property` getter method
- Enhanced error logging in `obstacle_callback` with `exc_info=True` for better debugging context
- Added comprehensive NumPy-style docstring to `__init__` method documenting initialization flow and thread safety mechanisms
- Updated class-level docstring to describe thread-safe characteristics and asynchronous callback behavior

**Technical Details:**
- The `obstacle_callback` method runs asynchronously in a separate thread when Zenoh receives point cloud data from the `camera/realsense2_camera_node/depth/obstacle_point` topic
- Without thread synchronization, concurrent access to `self.obstacle` from the callback thread and main thread could lead to race conditions
- The lock ensures atomic updates when processing point cloud data and calculating obstacle coordinates (x, y, z, angle, distance)

**Files Modified:**
- `src/providers/d435_provider.py`


Wanbogang (commented):
Can you please resolve the conflicts?

Wanbogang:
@xieNniu 
Thread locking is technically correct, but for this use case it adds complexity without significant benefit. Consider simplifying it.

PR #1521, opened 2026-01-16
fix: add VideoCapture release to prevent resource leak in VLM_Local_YOLO

0xbyt4:
## Summary

Fix resource leak in `check_webcam()` - VideoCapture was not being released.

## Problem

`check_webcam()` creates a temporary `cv2.VideoCapture` to check camera resolution but never releases it:

```python
def check_webcam(index_to_check):
    cap = cv2.VideoCapture(index_to_check)
    # ... checks resolution ...
    return width, height  # cap never released!
```

This locks the camera device unnecessarily.

## Solution

Added `cap.release()` before both return statements.


Wanbogang:
@0xbyt4 
This change actually adds unnecessary complexity.

0xbyt4:
> @0xbyt4 This change actually adds unnecessary complexity.

@Wanbogang  Without this, the temporary VideoCapture stays open and locks the camera device  and removed  complexity.

PR #1590, opened 2026-01-17
fix: correct inverted required logic and add error handling in schema generator

LordKaiRex:
Fixes

This PR fixes two bugs in the configuration schema generator that affect schema correctness and script reliability.

Changes Made:

1. Fixed Inverted 'required' Field Logic (CRITICAL)
`scripts/generate_config_schema.py` (Line 458)

Problem: Fields with default values were incorrectly marked as `required: True`, and fields without defaults were marked as `required: False`.

Before:
```python
"required": has_default,  ❌ BUG: Inverted logic
```

After:
```python
"required": not has_default,  ✅ Required when NO default
```

Impact: All generated schemas now correctly identify required vs optional fields.

---

2. Added Error Handling to `_py_files` Method
`scripts/generate_config_schema.py` (Lines 389-398)

Problem: Method would crash if directory doesn't exist.

Before:
```python
    return [
        os.path.join(directory, f)
        for f in os.listdir(directory)  # Could raise FileNotFoundError
        ...
    ]
```

After:
```python
        return []
    return [...]
```

Impact: Script handles missing directories gracefully.

Testing:

- ✅ Ran `pre-commit run --all-files`
- ✅ Verified required fields are now correctly identified
- ✅ Tested with missing directories to confirm graceful handling

Type of Change:

- [x] Bug fix (non-breaking change which fixes an issue)

Checklist:

- [x] My code follows the project's style guidelines
- [x] I have performed a self-review of my changes
- [x] My changes fix functional bugs, not stylistic issues
- [x] All pre-commit checks pass

### DIFF
--- a/scripts/generate_schema.py
+++ b/scripts/generate_schema.py
@@ -396,8 +396,12 @@ class ConfigSchemaGenerator:
         Returns
         -------
         List[str]
-            List of absolute paths to Python files.
+            List of absolute paths to Python files. Returns empty list if
+            directory does not exist.
         """
+        if not os.path.exists(directory):
+            return []
+
         return [
             os.path.join(directory, f)
             for f in os.listdir(directory)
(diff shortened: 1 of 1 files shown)


0xbyt4:
Hi, @LordKaiRex thanks for your PR ! Please split into separate PRs for unrelated changes and add regression tests for the bug fix.

PR #1617, opened 2026-01-18
fix: handle missing unitree sdk

dyhbrewer:
## Summary
This PR prevents the Unitree Camera provider from causing import/initialization errors in environments where the Unitree SDK is not installed.

## Changes
- Added a `try-except` block around the `VideoClient` initialization.
- catches `NameError` (which occurs when the SDK types/classes are not imported).
- Sets `self.video_client` to `None` if the SDK is missing.

## Motivation
The code previously assumed the presence of the Unitree SDK. In simulation or non-robot environments (like macOS testing), this caused the application to crash. This fix allows for graceful degradation.

### DIFF
--- a/src/providers/unitree_camera_vlm_provider.py
+++ b/src/providers/unitree_camera_vlm_provider.py
@@ -58,8 +58,13 @@ class UnitreeCameraVideoStream(VideoStream):
             jpeg_quality=jpeg_quality,
         )
 
-        self.video_client = VideoClient()  # type: ignore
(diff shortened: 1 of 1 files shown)


0xbyt4:
hi @dyhbrewer ,Thanks for identifying this issue, but I think this approach (silent failure) isn't the right solution.
If someone runs `unitree_go2_remote.json5` config, they expect the Unitree camera to work. Silent failure would hide the problem.

PR #1703, opened 2026-01-19
Fix bug tests

navyxie:
## Overview
Fix test collection failure caused by NameError when importing plugins with missing optional dependencies (e.g., unitree_g1_basic.py fails due to undefined dds_).

## Type of change
- [x] Bug fix

## Changes
- Extended exception handling in get_all_inputs_classes() function from (ImportError, ModuleNotFoundError) to (ImportError, ModuleNotFoundError, NameError, AttributeError)
- This ensures plugins that fail to import due to any type of missing optional dependency are gracefully skipped with a warning log, rather than crashing the entire test collection

## Checklist

- [x] Self-review completed
- [x] Documentation updated (if applicable)
- [x] Local testing completed
- [x] Tests added/updated (if applicable)

## Impact
- Before fix: Running pytest --collect-only would fail with NameError: name 'dds_' is not defined from unitree_g1_basic.py, causing test collection to abort
- After fix: Test collection completes successfully (308 tests collected), with problematic plugins logged as warnings and skipped

## Additional Information
The root cause is that unitree_g1_basic.py uses dds_.BmsState_ in a type annotation at class definition time (line 139), which raises NameError when the CycloneDDS dependency is not available. Unlike ImportError, this error was not being caught by the original exception handler.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,9 @@ wheels/
 # IDE
 .vscode/
 
+# Claude Code
+CLAUDE.md
+
 # Debug files
 front_image.jpg
 turtlebot4_debug.jpg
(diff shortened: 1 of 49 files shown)


navyxie:
Fix test collection failure caused by NameError when importing plugins with missing optional dependencies (e.g., unitree_g1_basic.py fails due to undefined dds_).

PR #1707, opened 2026-01-19
fix: reuse aiohttp ClientSession and rate-limit tracking updates

### LINKED ISSUE #1592: Critical bugs in person_follow mode (config parsing, lifecycle, async handling)
### 1. Invalid escape sequences in system_prompt_base (CRASH)
config/unitree_go2_modes.json5

The system_prompt_base contains invalid escape sequences like:
\BEHAVIOR, \1, \SAFETY

This can cause JSON5 or runtime parsing failures.
Only \n should be escaped.

---

### 2. Duplicate config key (silent overwrite)
config/unitree_go2_modes.json5

The key `history_length` is defined twice in the same config block.
This silently overwrites values and can cause unpredictable behavior.

---

### 3. Lifecycle hook executed twice
config/unitree_go2_modes.json5

`start_person_follow_hook` is called in both `on_startup` and `on_entry`,
causing double enrollment attempts.

---

### 4. aiohttp ClientSession recreated every poll (performance / leak risk)
src/inputs/plugins/person_following_status.py

A new `aiohttp.ClientSession()` is created on every poll call.
This can lead to socket exhaustion and performance degradation.

---

### 5. Tracking status message spam to LLM
src/inputs/plugins/person_following_status.py

While tracking is active, status messages are sent every poll cycle,
causing excessive LLM token usage.

---

### Suggested priority
Critical: 1–3  
Medium: 4–5

wtoqxy0000:
## Summary
- Fixes aiohttp ClientSession being recreated on every poll call, which can lead to socket exhaustion and unnecessary overhead
- Adds rate limiting for tracking status messages to prevent excessive LLM token usage (configurable via `tracking_update_interval`, default 5 seconds)

Fixes #1592 (bugs 4 and 5)

## Changes
- Added `_get_session()` method to lazily create and reuse `aiohttp.ClientSession`
- Added `tracking_update_interval` config parameter (default 5.0 seconds)
- Added `_last_tracking_update` state tracking
- Modified `_format_status()` to only return tracking updates periodically instead of every poll cycle

## Test plan
- [x] Syntax check passed
- [x] Ruff lint passed
- [x] Module import test passed
- [ ] Verify person following mode still functions correctly
- [ ] Monitor memory/socket usage to confirm no leaks


Wanbogang:
@wtoqxy0000 
Please take a look and fix the test errors

PR #1708, opened 2026-01-19
fix: add directory existence check in schema generator

### LINKED ISSUE #1653: Add error handling for missing directories in schema generator
Location: `scripts/generate_schema.py`, `_py_files` method (lines 389-398)

Problem:
The `_py_files` method calls `os.listdir(directory)` without checking if the directory exists first. This causes a `FileNotFoundError` when scanning missing directories.

Current code:
```python
    return [
        os.path.join(directory, f)
        for f in os.listdir(directory)  # ❌ Crashes if directory missing
        if f.endswith(".py") and f != "__init__.py"
    ]
```

Impact:
- Script crashes instead of handling missing directories gracefully
- Makes the tool fragile during development when directory structures change

Fix:
Add directory existence check before calling `os.listdir()`:
```python
    return []
```

I have a PR ready with the fix.

wtoqxy0000:
## Summary
- Added directory existence check in `_py_files` method to prevent FileNotFoundError when scanning missing directories

Fixes #1653

## Changes
- Added `os.path.exists()` check before `os.listdir()` in `_py_files` method
- Returns empty list gracefully when directory doesn't exist

## Test plan
- [x] Syntax check passed
- [x] Ruff lint passed
- [x] Local test passed


0xbyt4:
hi @wtoqxy0000  Every caller of _py_files() already checks os.path.exists() before calling it ,this fix adds a redundant check that can never be triggered.

PR #1710, opened 2026-01-19
fix: correct format string and add null check

wtoqxy0000:
## Summary
Two small bug fixes:

1. **vlm_local_yolo.py line 95**: Fix missing separator in resolution format string
   - Changed `{width}{height}` to `{width}x{height}`
   - Output now shows "1920x1080" instead of "19201080"

2. **fabric_map_provider.py line 229**: Add null check before `os.path.exists()`
   - Added `self.filename_current is not None` check
   - Matches the pattern used in `rplidar_provider.py` (line 313)
   - Prevents potential TypeError if filename_current is None

## Test plan
- [x] Syntax check passed
- [x] Ruff lint passed

### DIFF
--- a/src/inputs/plugins/vlm_local_yolo.py
+++ b/src/inputs/plugins/vlm_local_yolo.py
@@ -92,7 +92,7 @@ def check_webcam(index_to_check):
 
     # Set the best available resolution
     width, height = set_best_resolution(cap, RESOLUTIONS)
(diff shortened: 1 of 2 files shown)


0xbyt4:
Hi @wtoqxy0000, please split your PRs ,  one fix per PR.

wtoqxy0000:
OK, thanks.

PR #1711, opened 2026-01-19
fix: Correct inverted required field logic in schema generator

### LINKED ISSUE #1610: Critical bug: Inverted 'required' field logic in schema generator
Location: `tests/test_generate_config_schema.py`, line 458

Problem:
The schema generator incorrectly sets `"required": has_default`, which is backwards. A field should be marked as required when it does NOT have a default value.

Current code:
```python
field = {
    "name": name,
    "type": self._annotation_to_type(item.annotation),
    "label": name.replace("_", " ").title(),
    "required": has_default,  # ❌ BUG: Inverted logic
    "description": description,
}
```

Current behavior:
- Fields WITH defaults are marked as required ❌
- Fields WITHOUT defaults are marked as optional ❌

Expected behavior:
- Fields WITH defaults should be optional ✅
- Fields WITHOUT defaults should be required ✅

Impact:
All generated schemas have inverted validation rules. This causes:
- Optional fields to appear as required
- Required fields to appear as optional
- Schema validation failures

Fix:
Change line 458 from `"required": has_default,` to `"required": not has_default,`

I have a PR ready with the fix.

nan1888:
Fixes #1610

Problem:
- Schema generator had inverted logic for 'required' field
- Fields WITH defaults were marked as required
- Fields WITHOUT defaults were not marked as required

Solution:
- Changed `required: has_default` to `required: not has_default`

Impact:
- Config schema will now correctly identify required fields


Wanbogang (approved):
**Bug confirmed:** Inverted logic in required field detection.

**Fix verified:**
- Before: `required = has_default`  (fields with defaults marked required)
- After: `required = not has_default`  (correct logic)

**Logic validated:**
- Field without default → required = True 
- Field with default → required = False

Wanbogang (approved):

PR #1713, opened 2026-01-19
fix: Add directory existence check in schema generator

### LINKED ISSUE #1589: Fix critical bugs in config schema generator
I found two bugs in `scripts/generate_schema.py` that affect schema generation:

 Bug 1: Inverted 'required' field logic (CRITICAL)
Location: Line 458 in `_parse_pydantic_fields_from_node` method

Problem:
The code sets `"required": has_default`, which is backwards. A field should be required when it does NOT have a default value.

Current behavior:
- Fields with defaults appear as required ❌
- Fields without defaults appear as optional ❌

Expected behavior:
- Fields with defaults should be optional ✅
- Fields without defaults should be required ✅

Impact: All generated schemas have inverted validation rules, causing schema validation to fail or behave incorrectly.

---

Bug 2: Missing error handling in `_py_files` method
Location: Lines 389-398

Problem:
The method calls `os.listdir(directory)` without checking if the directory exists, causing `FileNotFoundError` when directories are missing.

Impact: Script crashes instead of gracefully handling missing directories.

---

I have a PR ready with both fixes once approved.

nan1888:
Fixes #1589

Problem:
- _py_files() crashes when directory doesn't exist

Solution:
- Added os.path.exists() check

Impact:
- Graceful handling of missing directories

### DIFF
--- a/.github/ISSUE_TEMPLATE/🌟-bounty.md
+++ b/.github/ISSUE_TEMPLATE/🌟-bounty.md
@@ -14,7 +14,7 @@ assignees: ''
 **Guidelines**
 
 **License**
-This project is licensed under the MIT License. See the [LICENSE](https://github.com/OpenmindAGI/OM1/blob/main/LICENSE) file for more details.
+This project is licensed under the MIT License. See the [LICENSE](https://github.com/Openmind/OM1/blob/main/LICENSE) file for more details.
 
 **Contact**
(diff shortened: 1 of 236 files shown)


scripts/generate_schema.py
Wanbogang, line 541: "required": not has_default, 

please change

PR #1727, opened 2026-01-19
fix(providers): add resource management to RtkProvider stop method ✨

yimu0080:
Hi! 👋 I noticed that the `RtkProvider.stop()` method wasn't closing the serial connection, which could lead to resource leaks. So I added proper cleanup logic! ✨

**Changes:**
- Added serial connection cleanup in `stop()` method to prevent resource leaks
- Enhanced error logging with context (port name) for better debugging
- Updated docstring to document the cleanup process

The fix ensures that when `RtkProvider.stop()` is called, the serial connection is properly closed if it's open, similar to how `GpsProvider` handles it. This prevents resource leaks and makes debugging easier with better error messages! 🚀


Wanbogang:
@yimu0080 
Please fix the lint/test failures in the CI pipeline

PR #1728, opened 2026-01-19
fix(providers): add resource cleanup to config_provider stop method

yixing0098:
added exception handling and finally block to ConfigProvider.stop() method.

ensures session, config_response_publisher, and config_request_subscriber
are properly cleaned up even if errors occur during shutdown. added
exc_info=True logging for better debugging.


0xbyt4:
hi @yixing0098  ,Thanks for contribution!
 `session.close()` doesn't raise exceptions. Also, the current `ZError: session closed` message is more descriptive than `AttributeError: NoneType` for debugging race conditions.

PR #1784, opened 2026-01-20
fix(providers): remove try-except from ContextProvider stop method

yixing0098:
Removed try-except block from ContextProvider.stop() method to comply with project rules.

Changes:
- Removed try-except wrapper around session.close() call
- Preserved resource cleanup logic (session close, reference clearing)
- Updated docstring to document the 3-step cleanup process
- Added Notes section explaining exception propagation mechanism

This change allows underlying exceptions to propagate, enabling maintainers to debug race conditions. Resource cleanup (session and publisher reference clearing) is preserved.

### DIFF
--- a/src/providers/context_provider.py
+++ b/src/providers/context_provider.py
@@ -76,13 +76,20 @@ class ContextProvider:
     def stop(self):
         """
         Stop the ContextProvider and clean up resources.
+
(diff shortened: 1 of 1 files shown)


Wanbogang:
@yixing0098 Breaks project rules.Thanks

PR #1837, opened 2026-01-22
fix: add URL validation for Slack webhook in nightly integration tests

aydnOktay:
## Overview
Fixes the nightly integration tests workflow failure caused by missing `ONCALL_SLACK_WEBHOOK_URL` secret. The workflow was failing with "curl: (2) no URL specified" error when trying to send Slack notifications, causing continuous workflow failures on scheduled runs.

## Problem
The workflow was attempting to send Slack notifications without checking if the `ONCALL_SLACK_WEBHOOK_URL` secret exists. When the secret is not configured, curl fails with exit code 2, causing the entire workflow to fail. This has been causing continuous workflow failures on the main branch's scheduled runs (every hour), resulting in notification spam.

**Error observed:**

## Solution
- Added secret existence check in the workflow condition (`secrets.ONCALL_SLACK_WEBHOOK_URL != ''`)
- Added runtime URL validation in the shell script to gracefully skip notification if secret is empty
- Wrapped the URL in quotes to handle edge cases
- Applied fix to both failure and success notification steps

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation improvement
- [ ] Other

## Changes
- `.github/workflows/nightly-integration-tests.yml`:
  - Added secret check in `if` condition for both notification steps
  - Added runtime URL validation with graceful exit
  - Wrapped webhook URL in quotes for safety

## Impact
- **Before**: Workflow fails with exit code 2 when Slack webhook secret is not configured, causing continuous failures on scheduled runs (every hour)
- **After**: Workflow continues successfully, skipping Slack notifications if secret is not available
- No breaking changes - existing behavior preserved when secret is configured

## Testing
- Verified workflow syntax is valid
- Tested that secret check condition works correctly
- Workflow will now gracefully skip notifications instead of failing

### DIFF
--- a/.github/workflows/nightly-integration-tests.yml
+++ b/.github/workflows/nightly-integration-tests.yml
@@ -123,8 +123,12 @@ jobs:
           EOF
 
       - name: Notify Slack on failure
-        if: failure() && github.event_name == 'schedule'
+        if: failure() && github.event_name == 'schedule' && secrets.ONCALL_SLACK_WEBHOOK_URL != ''
         run: |
+          if [ -z "${{ secrets.ONCALL_SLACK_WEBHOOK_URL }}" ]; then
+            echo "ONCALL_SLACK_WEBHOOK_URL secret is not set, skipping Slack notification"
+            exit 0
+          fi
           curl -X POST -H 'Content-type: application/json' \
             --data "{
               \"text\": \"❌ @channel *Integration Tests Failed*\n\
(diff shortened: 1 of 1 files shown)


aydnOktay:
This is a bug fix. The workflow keeps failing and I keep receiving email notifications. If I don't need to touch the .github/ folder, please suggest an alternative solution @openminddev

PR #1904, opened 2026-01-26
feat: Persistent JSON Storage for Conversation History (#985)

farukk77:
Implemented a JSON-based storage mechanism to save and restore conversation history upon restart, resolving the issue of data loss during crashes.

### DIFF
--- a/docs/api-reference/api_pricing.mdx
+++ b/docs/api-reference/api_pricing.mdx
@@ -1,0 +1,86 @@
+---
+title: Subscription Plans
+description: "API Pricing"
+---
+
+The APIs now support a new unit for billing, **OMCU (OpenMind Computational Unit)**. See details below for our free, standard, builder, pro, and enterprise plans.
+
+Note: Each subscription plan includes all features offered in the lower-tier plans
+
+<Card title="Free" icon="coin">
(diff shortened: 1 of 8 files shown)


farukk77:
Update: Refactor to Asynchronous Non-Blocking I/O
Upon further review and aligning with the maintainers' request for a "simple JSON solution", I have upgraded the implementation to ensure high performance.
I identified that standard synchronous file writes (open().write) could block the main event loop, causing the robot to "freeze" during history saves. To fix this, I refactored the solution to be fully asynchronous.
Key Improvements in this Update:
Non-Blocking Performance: Switched to aiofiles and asyncio.create_task. Saving now happens in the background ("Fire & Forget"), ensuring the runtime remains 100% responsive.
Atomic & Crash-Safe: Implemented a write to temp -> os.replace strategy. This prevents JSON corruption if the system crashes mid-write.
Maintained Simplicity: The storage format remains a standard, human-readable JSON list, as requested.
Files Changed:
requirements.txt: Added aiofiles.
src/memory/async_json_storage.py: Created for low-level async handling.
src/providers/llm_history_manager.py: Integrated the new async backend.

PR #1937, opened 2026-01-27
perf(unitree): decouple websocket I/O from video capture loop

MuhamadRifansyah:
## Overview
This PR improves the runtime robustness of the UnitreeRealSenseDevVideoStream by decoupling the video capture loop from the network transmission logic. Previously, network callbacks (WebSocket sends) were executed synchronously within the video capture loop, causing frame rate drops and potential blocking if the network was slow or unstable.

## Changes
[src/providers/unitree_realsense_dev_vlm_provider.py](code-assist-path:c:\Users\asus\Music\om\OM1\src\providers\unitree_realsense_dev_vlm_provider.py)
Architecture: Implemented a Producer-Consumer pattern.
Added a queue.Queue (maxsize=2) to buffer encoded frames.
Added a dedicated background thread (_send_loop) to consume frames from the queue and execute callbacks.
Logic:
Updated on_video to push frames to the queue using put_nowait. If the queue is full (network lag), the oldest frame is effectively skipped to maintain real-time latency.
Updated start() and stop() methods to manage the lifecycle of the new sender thread.

## Impact
Performance: The video capture loop now runs independently of network latency. High-latency network conditions will no longer slow down the camera read operations.
Robustness: Prevents the entire provider from hanging if a WebSocket send operation blocks.
Latency: Ensures that the system always processes the most recent frames available, dropping older frames if the consumer cannot keep up.

## Testing
[x] Verified that start() initializes both the capture and sender threads.
[x] Verified that stop() cleanly joins the sender thread and stops execution.
[x] Confirmed that frames are still received by the WebSocket client.
[x] Validated that the application does not crash if the queue fills up (frames are dropped gracefully).

## Additional Information
This addresses the "Synchronous Network I/O in Real-Time Loop" finding identified during the runtime robustness review.


Wanbogang (approved):
Decoupling the video capture loop from network I/O via a producer–consumer pattern improves robustness and real-time performance

PR #1955, opened 2026-01-27
 Add serial connection cleanup to RtkProvider.stop

aydnOktay:
While reviewing the codebase, I noticed that RtkProvider.stop() wasn't closing the serial connection, which could cause resource leaks when the provider is stopped. GpsProvider already handles this cleanup, so I added the same serial port closure logic to RtkProvider.stop() for consistency and proper resource management. The fix checks if the serial connection is open before closing it and includes error handling with logging, matching the pattern used in GpsProvider. This ensures resources are cleaned up properly when the provider stops.


aydnOktay:
@openminddev

aydnOktay:
@openminddev

PR #1964, opened 2026-01-28
cli: prepare follow-up to relax AST-based class validation

MuhamadRifansyah:
The CLI currently verifies component existence using AST parsing,
requiring the class to be defined at the top level of the file.

This approach can produce false negatives for valid plugins that
re-export classes or use indirection, leading to confusing developer
errors.

This follow-up PR will improve developer experience by either relaxing
the validation or clearly documenting its limitations, without changing
runtime behavior.

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -586,6 +586,9 @@ def _check_class_in_dir(directory: str, class_name: str) -> bool:
     """
     Check if a class exists in any .py file in the given directory using AST.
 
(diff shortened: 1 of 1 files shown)


0xbyt4:
Hi @MuhamadRifansyah , Import check passes syntax validation but doesn't verify the module actually exists. Original ClassDef-only was more conservative  only reports classes that  are definitely defined in the file.

PR #2006, opened 2026-01-30
Add error handling to ZenohListenerProvider.stop

aydnOktay:
While reviewing provider cleanup methods, I noticed that **ZenohListenerProvider.stop()** was calling **session.close()** without any exception handling, which could cause the provider to crash if the session close operation fails. I also noticed the docstring incorrectly mentioned stopping a background thread when no thread actually exists in this provider. I've added proper try-except-finally error handling similar to **ContextProvider** , ensuring that exceptions during session closure are caught and logged, and the session is always set to None in the finally block for proper cleanup. The docstring has also been corrected to accurately reflect what the method does


0xbyt4:
Hi @aydnOktay  can you fix  lint errors

aydnOktay:
> Hi @aydnOktay can you fix lint errors

done sir , check pls @0xbyt4  @openminddev

PR #2008, opened 2026-01-30
fix(config): validate config structure before read and write

MuhamadRifansyah:
## Overview
This PR improves the robustness of the ConfigProvider by enforcing strict type validation on configuration data. It ensures that configuration payloads—whether loaded from disk or received via Zenoh updates—are valid dictionary structures, preventing invalid JSON types (such as lists or primitives) from corrupting the runtime state or causing downstream crashes.

## Changes
[src/providers/config_provider.py](code-assist-path:c:\Users\asus\Music\om\OM1\src\providers\config_provider.py):
Added explicit type validation in _handle_set_config to ensure incoming configuration updates are dictionaries before writing them to disk.
Added type validation in _get_config_snapshot to verify that configuration loaded from the file system is a dictionary.
Implemented clear error logging and exception raising when invalid configuration structures are detected.

## Impact
Stability: Prevents runtime errors that occur when downstream components expect a dictionary interface (e.g., .get()) but receive a list or primitive.
Data Integrity: Protects the persistent runtime configuration file from being overwritten with invalid data structures.
Observability: Provides specific error messages identifying the received type versus the expected type, making misconfiguration easier to debug.

## Testing
Invalid Update Test: Send a Zenoh ConfigRequest containing a JSON list (e.g., "[1, 2]") or primitive. Verify that the provider raises a ValueError and does not overwrite the config file.
Corrupt File Test: Manually modify .runtime.json5 to contain a JSON list. Restart the provider and verify it logs an error and returns an empty dictionary instead of crashing.
Valid Flow: Verify that standard dictionary configurations continue to load and update correctly.

## Additional Information
This change addresses a potential failure mode where json5 successfully parses valid JSON that is structurally invalid for our application's configuration schema (e.g., a top-level list).

### DIFF
--- a/src/providers/config_provider.py
+++ b/src/providers/config_provider.py
@@ -104,6 +104,11 @@ class ConfigProvider:
         try:
             new_config = json5.loads(config_str)
 
+            if not isinstance(new_config, dict):
+                raise ValueError(
+                    f"Invalid configuration structure: Expected 'dict', got '{type(new_config).__name__}'"
+                )
+
             temp_path = self.config_path + ".tmp"
             with open(temp_path, "w") as f:
                 json.dump(new_config, f, indent=2)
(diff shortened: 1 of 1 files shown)


0xbyt4:
Hi @MuhamadRifansyah  Thanks for Contribution! PR removes undeclare() calls from stop() causing resource leak.

PR #2039, opened 2026-02-01
fix(docs): correct parameter type in zenoh

windlgrass:
Fixed incorrect parameter type documentation in `_handle_zenoh_message`

Changed `zenoh_input : object` to `zenoh_input : zenoh.Sample` to match the actual type hint


Wanbogang:
@windlgrass 
Thanks for the contribution attempt, but let's focus on more impactful changes!

PR #2049, opened 2026-02-01
Warn when action_execution_mode/action_dependencies are defined but unused

Kriptoboss:
### What
- Emit warnings when `action_execution_mode` and/or `action_dependencies` are set in config but not yet consumed by the runtime.

### Why
These fields are currently accepted by config validation but silently ignored at runtime, which can mislead users into thinking they are effective.

### Tested
- `uv run src/cli.py validate-config config/spot_actiondeps_test.json5`
- `PYTHONPATH=src uv run python3 -c "from runtime.single_mode.config import load_config; load_config('spot_actiondeps_test')"`

### DIFF
--- a/src/runtime/multi_mode/config.py
+++ b/src/runtime/multi_mode/config.py
@@ -394,6 +394,17 @@ def load_mode_config(
     verify_runtime_version(config_version, config_name)
     validate_config_schema(raw_config)
 
+    # Warn if config fields are defined but not yet consumed by the runtime.
+    if raw_config.get("action_execution_mode"):
+        logging.warning(
+            "action_execution_mode is defined but not yet consumed by the runtime (config-only for now)."
+        )
+    if raw_config.get("action_dependencies"):
+        logging.warning(
+            "action_dependencies are defined but not yet consumed by the runtime (config-only for now)."
+        )
+
+
     g_robot_ip = raw_config.get("robot_ip", None)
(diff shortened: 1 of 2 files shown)


Kriptoboss:
### Tested
- uv run src/cli.py validate-config config/spot_actiondeps_test.json5
- PYTHONPATH=src uv run python3 -c "from runtime.single_mode.config import load_config; load_config('spot_actiondeps_test')"

Kriptoboss:
Hi! 👋  
It looks like only the PR labeler ran successfully.  
The remaining CI checks (lint, unit tests, package install, nightly) seem to be pending on this fork PR.

Could a maintainer please approve or re-run the workflows so the remaining checks can execute?

Thanks!

0xbyt4:
Hi @Kriptoboss Thank you for contribution!  These fields are already consumed by ActionOrchestrator  so these warnings would be incorrect.

Kriptoboss:
Thanks for the clarification, that makes sense 👍

From a contributor’s perspective, it wasn’t obvious that these fields are
already consumed by ActionOrchestrator, which is why they appeared unused
at first glance.

I agree the warnings would be incorrect in this case. Happy to close or
adjust this PR — and it might still be worth considering some form of
documentation or comment to make this behavior clearer for future
contributors.

Appreciate the explanation!

PR #2061, opened 2026-02-02
fix(tests): implement test_poll_returns_none_if_requests_is_none_and_mock_disabled

LupaFlex:
## Problem

The test `test_poll_returns_none_if_requests_is_none_and_mock_disabled` only contained `pass` - it was not implemented and didn't test anything.

## Solution

Implemented the test to verify that `_poll()` returns `None` when:
- `requests` module is unavailable (set to `None`)
- Mock mode is disabled

## Changes

- `tests/inputs/plugins/test_fabric_closest_peer.py`: Implemented the empty test with actual test logic

## Testing

- Pre-commit checks pass
- Test follows existing patterns in the file


0xbyt4:
Hi @LupaFlex Thanks for contribution! This test covers an unreachable code path.

PR #2063, opened 2026-02-02
docs: clarify action fields are consumed by ActionOrchestrator

Kriptoboss:
Adds a small clarification in an example config to prevent future contributors from assuming action_execution_mode/action_dependencies are unused.

No runtime behavior changes — comment only.


Kriptoboss:
Context: this is a small follow-up to #2049, adding a clarification so these fields don’t appear unused to future contributors.

No behavior change intended.

Issue #1167, opened 2026-01-09
Add WSL Setup Guide for Windows Users


OxHanta:
## Problem
Currently, there's no dedicated guide for Windows users who want to run OM1 via WSL. 
This creates a barrier for Windows developers interested in deploying OM1 robots.

## Proposed Solution
Add a comprehensive WSL setup guide that includes:
- WSL2 installation prerequisites
- Step-by-step OM1 deployment on WSL

## Benefits
- Lowers the barrier to entry for Windows users
- Provides clear, tested instructions for WSL deployment
- Includes troubleshooting for common WSL-specific issues

I've already created this guide and tested it: https://github.com/OxHanta/OpenMind-OM1-WSL-Guide

Happy to submit a PR if this aligns with the project's documentation goals.

omerbek:
I’d like to work on adding a WSL setup guide for Windows users.

Prachi1615:
Hi @OxHanta 
Thanks for raising this.
We don't plan to add support for windows. 
Closing this issue now.

PR #2076, opened 2026-02-03
fix(config): make runtime config updates Windows-safe and explicit on…

MuhamadRifansyah:
# Overview
Fixes `ConfigProvider` runtime config updates on Windows and prevents silent “empty config” responses when the runtime config file is missing or unreadable.

# Changes
- Switch from `os.rename()` to `os.replace()` for atomic overwrites of `config/memory/.runtime.json5` on Windows (`src/providers/config_provider.py:111`).
- Remove `{}` fallback in `_get_config_snapshot()` so missing/unreadable configs trigger the existing error-response path instead of returning an empty snapshot (`src/providers/config_provider.py:164`).
- Add targeted unit tests for overwrite behavior and missing-config error response (`tests/providers/test_config_provider.py:35`, `tests/providers/test_config_provider.py:56`).

# Impact
- Windows: “set config” updates work even when the destination file already exists.
- “get config” no longer returns `{}` with a success message on missing/corrupt config; it returns an error response with a clear message.
- No schema or CI workflow changes.

# Testing
- Added `tests/providers/test_config_provider.py`.
- Local test run was blocked on Windows because `uv run pytest` fails to resolve/install `tensorflow-io-gcs-filesystem==0.37.1` (no compatible Windows wheel in current resolution).

# Additional Information
This change preserves runtime behavior for valid configs and only improves correctness and diagnostics in failure modes.

### DIFF
--- a/src/providers/config_provider.py
+++ b/src/providers/config_provider.py
@@ -108,7 +108,7 @@ class ConfigProvider:
             with open(temp_path, "w") as f:
                 json.dump(new_config, f, indent=2)
 
-            os.rename(temp_path, self.config_path)
+            os.replace(temp_path, self.config_path)
 
             logging.info(f"Updated runtime config file: {self.config_path}")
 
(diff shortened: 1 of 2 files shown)


0xbyt4:
Hi @MuhamadRifansyah, thanks for the contributions! However, OM1 doesn't officially support Windows (see #1167). Closing #2076, #2104, and #2115.

PR #2097, opened 2026-02-04
ix: Strict ABI Validation to Resolve Ambiguity (Fixes #1824)

### LINKED ISSUE #1824: [BUG] Potential Bug in `decode_eth_response` in `src/inputs/plugins/ethereum_governance.py`

farukk77:
This PR addresses **Issue #1824** regarding the ambiguity in parsing Ethereum RPC responses.

The previous manual slicing approach caused "Silent Failures" because it could not distinguish between valid empty data and corrupted bytes.

### The Fix
I have refactored `decode_eth_response` to use the **`eth-abi`** library for strict validation instead of manual string manipulation.

### How it handles the Scenarios (as defined in #1824):

1.  **Scenario A (Valid Empty):**
    * The code explicitly checks for `"0x"` and returns an empty string. This is treated as the only valid "no data" state.

2.  **Scenario B (Network/RPC Error):**
    * Checks if the input is `None` (RPC failure) and logs a specific warning instead of crashing or returning empty.

3.  **Scenario C (Corrupted/Truncated Data):**
    * **Old Behavior:** Silently returned `""` (indistinguishable from Scenario A).
    * **New Behavior:** The `eth_abi.decode` function raises a `DecodingError` if the bytes are malformed. I capture this exception and log a critical error.

### Verification
This implementation ensures that network glitches or corrupted bytes are flagged explicitly, eliminating the ambiguity for the governance listener.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,2 @@
+eth-abi>=4.0.0
+eth-utils>=2.0.0
--- a/src/inputs/plugins/ethereum_governance.py
+++ b/src/inputs/plugins/ethereum_governance.py
@@ -1,51 +1,35 @@
-import asyncio
 import logging
-import time
+import aiohttp
 from typing import Optional
 
-import aiohttp
+# PR #2069: Use strict decoding library instead of manual slicing
+from eth_abi import decode
+from eth_utils import to_bytes
+from eth_abi.exceptions import DecodingError
 
-from inputs.base import Message, SensorConfig
+from inputs.base import SensorConfig
 from inputs.base.loop import FuserInput
-from providers.io_provider import IOProvider
 
-# RULES are stored on the ETHEREUM HOLESKY testnet
-# "GovernanceContract": "0xe706b7e30e378b89c7b2ee7bfd8ce2b91959d695"
-# https://holesky.etherscan.io/address/0xe706b7e30e378b89c7b2ee7bfd8ce2b91959d695
-# Note: Etherscan.io does not handle bytes[]/json well. See below for ways to
-# interact with HOLESKY and decode the data, generating an ASCII string.
-
+logger = logging.getLogger(__name__)
 
 class GovernanceEthereum(FuserInput[SensorConfig, Optional[str]]):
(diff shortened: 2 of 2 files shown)


0xbyt4:
Hi @farukk77  Thanks for contribution! you  remove most of the class (including __init__ and _poll) to fix one method , it would break the governance system entirely, and new deps belong in  pyproject.toml not requirements.txt.

PR #2115, opened 2026-02-05
fix(io): read config files with explicit UTF-8 encoding

MuhamadRifansyah:
**# Overview**
- Standardize UTF‑8 decoding for config and schema reads to avoid platform‑dependent parsing issues.

**# Changes**
- Add `encoding="utf-8"` to JSON/JSON5 file opens in `src/runtime/config.py`, `src/runtime/single_mode/config.py`, `src/runtime/multi_mode/config.py`, `src/cli.py`, and `src/run.py`.

**# Impact**
- Prevents Windows/default‑encoding surprises while preserving behavior for valid UTF‑8 configs.

**# Testing**
- Not run (not requested).

**# Additional Information**
- No schema or runtime behavior changes; this is a read‑time safeguard only.

### DIFF
--- a/src/cli.py
+++ b/src/cli.py
@@ -199,7 +199,7 @@ def validate_config(
 
         # Load and parse JSON5
         try:
-            with open(config_path, "r") as f:
+            with open(config_path, "r", encoding="utf-8") as f:
                 raw_config = json5.load(f)
         except ValueError as e:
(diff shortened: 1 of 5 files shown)


0xbyt4:
Hi @MuhamadRifansyah   OM1 doesn't officially support Windows,  there are a lots  other  missing places  too with same pattern .

PR #2116, opened 2026-02-05
fix(llm): improve diagnostics for swallowed exceptions

MuhamadRifansyah:
**# Overview**
- Improve error diagnostics for OpenAI LLM calls and function schema generation without changing behavior.

**# Changes**
- Replace message-only logging with traceback-aware logging in `src/llm/plugins/openai_llm.py` and `src/llm/function_schemas.py`.

**# Impact**
- Makes root causes visible in logs while preserving current return values and control flow.

**# Testing**
- Not run (not requested).

**# Additional Information**
- No new dependencies, no schema changes, no runtime behavior changes.

### DIFF
--- a/src/llm/function_schemas.py
+++ b/src/llm/function_schemas.py
@@ -111,8 +111,8 @@ def generate_function_schemas_from_actions(actions: list) -> list[dict]:
                     f"Generated function schema for {action.llm_label}: {schema}"
                 )
             except Exception as e:
-                logging.error(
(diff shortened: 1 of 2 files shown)


0xbyt4:
Hi @MuhamadRifansyah, thanks for the improvement! However, this logging.error → logging.exception pattern exists across the entire codebase  changing only 2   files increases inconsistency ?

PR #2150, opened 2026-02-07
fix(config): validate required robot connection settings at startup

MuhamadRifansyah:
# Overview
Fail fast at startup when robot-dependent components are configured without a valid robot IP, preventing silent misconfiguration and late-stage runtime crashes.

# Problem
`robot_ip` defaults to `None` and previously only emitted warnings when missing.  
This allowed the system to start successfully, but caused failures deep in connector logic with unclear errors.

# Solution
Add explicit startup validation to detect when robot-dependent components are present but `robot_ip` is missing or invalid, and raise a clear `ValueError` with actionable diagnostics.

# How It Works
**Before**
- Missing `robot_ip` → warning logged
- System starts
- Crash occurs later during action execution

**After**
- Missing/invalid `robot_ip` + robot component present
- Clear startup failure with detailed error message
- No ambiguous runtime crashes

# Validation Logic
The startup validation:
- ✅ Passes when `robot_ip` is valid
- ✅ Passes when no robot-dependent components are configured (simulator-only setups)
- ❌ Fails fast when robot components are present but `robot_ip` is missing or invalid

Detected robot-dependent component types include:
- Unitree (`UnitreeGo2*`, `UnitreeG1*`, `UnitreeH1*`, `UnitreeB2*`)
- UBTech (`UbTech*`, `YansheeAPI`)
- Any component whose type matches these prefixes

# Error Message Example
```text
ValueError: Configuration requires robot connectivity but 'robot_ip' is missing or invalid.

Robot-dependent components found:
  input 'odom' (type: UnitreeGo2Odom)
  action 'move' (type: UnitreeGo2Move)

Please provide a valid robot IP address by:
  1. Adding 'robot_ip: "<robot-ip-address>"' to your config file, OR
  2. Setting the ROBOT_IP environment variable

### DIFF
--- a/src/runtime/single_mode/config.py
+++ b/src/runtime/single_mode/config.py
@@ -19,6 +19,70 @@ from simulators import load_simulator
 from simulators.base import Simulator
 
 
+def _check_robot_config_requirements(raw_config: dict) -> None:
+    """
+    Validate that robot_ip is present when robot-dependent components are configured.
+
+    Checks for components that require robot connectivity:
+    - Unitree inputs/actions/simulators
+    - UBTech Yanshee inputs/actions
+    - Boston Dynamics Spot actions
+    - Any component prefixed with 'Unitree' or 'UBTech'
+
+    Raises
+    ------
+    ValueError
+        If robot-dependent components are present but robot_ip is missing or invalid.
+    """
+    robot_ip = raw_config.get("robot_ip")
+    # Check if robot_ip is missing or invalid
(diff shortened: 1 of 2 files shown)


0xbyt4:
`"UbTech" in "UbtechASRInput"` and `"YansheeAPI" in "MoveYansheeConnector"` both return False due to case mismatch, so all UBTech/Yanshee components bypass this validation  , consider using case-insensitive matching or actual class name prefixes

PR #2155, opened 2026-02-07
Add tests for action loading system (describe_action and load_action) 

DasRasyo:
## Problem

  The `actions/__init__.py` module contains `describe_action` and `load_action` — core functions responsible for dynamically
   loading action plugins and generating LLM-facing descriptions. These functions currently have no test coverage, making it
   easy for regressions to go unnoticed.

  ## Changes

  Added `tests/actions/test_init.py` with 12 tests covering:

  **`describe_action` (6 tests):**
  - Actions with `exclude_from_prompt=True` return empty string
  - Enum fields list allowed values in the description
  - String fields show type annotations
  - Interface docstrings are included
  - Modules without Interface subclass return empty string
  - Non-existent modules return empty string

  **`load_action` (6 tests):**
  - Successful load returns correct AgentAction
  - Custom config classes from connector modules are used
  - Missing connector raises ValueError
  - Missing interface raises error
  - Default `exclude_from_prompt` is False
  - `exclude_from_prompt=True` is passed through correctly

  ## Test plan

  - [x] All new tests pass with `pytest tests/actions/test_init.py`
  - [x] Follows existing test patterns in the repository
  - [x] No changes to source code

### DIFF
--- a/tests/actions/test_init.py
+++ b/tests/actions/test_init.py
@@ -1,0 +1,233 @@
+  from dataclasses import dataclass
+  from enum import Enum
+  from types import ModuleType
+  from unittest.mock import patch
+
+  import pytest
+
+  from actions import describe_action, load_action
+  from actions.base import ActionConfig, ActionConnector, AgentAction, Interface
+
+
+  class SampleEnum(str, Enum):
+      OPTION_A = "option_a"
+      OPTION_B = "option_b"
+
+
+  @dataclass
+  class EnumInput:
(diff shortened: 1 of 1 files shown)


0xbyt4:
Hey @DasRasyo, thanks for adding test! However, there are several issues 
  IndentationError - file doesn't parse                                                                                                                        
  describe_action is called with wrong signature
  Wrong expected return values
  load_action tests missing connector key

PR #2156, opened 2026-02-07
fix:log silent exception in Qwen tool call parser 

DasRasyo:
## Overview                                                                                       
  The `_parse_qwen_tool_calls` function catches JSON parsing errors silently with a bare `except
  Exception: continue`, making it difficult to debug why tool calls are being dropped.

  ## Type of change
  - Bug fix

  ## Changes
  - `src/llm/plugins/qwen_llm.py` - Qwen tool call parsing errors were silently skipped. Added
  `logging.warning` so failed tool call blocks are visible in logs.

  ## Checklist
  - [x] Self-review completed
  - [x] Local testing completed

  ## Impact
  No behavioral changes. Only adds a log message to a previously silent exception handler, improving
  debuggability without affecting runtime flow.

### DIFF
--- a/src/llm/plugins/qwen_llm.py
+++ b/src/llm/plugins/qwen_llm.py
@@ -51,7 +51,8 @@ def _parse_qwen_tool_calls(text: str) -> list:
                         },
                     }
                 )
-        except Exception:
+        except Exception as e:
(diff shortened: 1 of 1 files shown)


Wanbogang:
@DasRasyo #2103 
Before you open a PR, it's a good idea to check it first to avoid duplication. Thanks for your contribution.

PR #2177, opened 2026-02-09
Refactor: Inject TTS provider into Unitree navigation providers for better testability

Budalebah:
## Problem
The `UnitreeGo2NavigationProvider` and `UnitreeG1NavigationProvider` classes were directly instantiating their TTS (Text-to-Speech) dependencies internally. This tight coupling made it difficult to mock the TTS service during unit testing, leading to potential side effects (like actual audio output) or failures in test environments where audio devices are unavailable.

## Solution
- Refactored `UnitreeGo2NavigationProvider` and `UnitreeG1NavigationProvider` to accept an optional `tts_provider` argument in their `__init__` methods (Dependecy Injection).
- Updated the initialization logic to use the injected provider if available; otherwise, it falls back to the default implementation as before.
- This change allows tests to inject a mock TTS provider, isolating the navigation logic from the TTS implementation.

## Type of Change
- [x] Refactor (no new functionality, no behavior change)

## Testing
- Existing tests should pass as the default behavior remains unchanged.
- New unit tests can now be written with mocked TTS providers.

### DIFF
--- a/src/providers/unitree_g1_navigation_provider.py
+++ b/src/providers/unitree_g1_navigation_provider.py
@@ -49,6 +49,7 @@ class UnitreeG1NavigationProvider:
         navigation_status_topic: str = "navigate_to_pose/_action/status",
         goal_pose_topic: str = "goal_pose",
         cancel_goal_topic: str = "navigate_to_pose/_action/cancel_goal",
+        tts_provider: Optional[object] = None,
     ):
         """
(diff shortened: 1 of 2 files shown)


0xbyt4:
Hi @Budalebah, the existing tests already mock ElevenLabsTTSProvider using unittest.mock.patch, which achieves the same isolation without changing the public API. See  other tests for examples.

PR #2178, opened 2026-02-09
Feat: Auto-create Coinbase Wallet if ID is missing

Budalebah:
## Problem
Currently, the `WalletCoinbase` plugin raises an initialization error if `COINBASE_WALLET_ID` is missing from the configuration, even when valid API keys (`COINBASE_API_KEY` and `COINBASE_API_SECRET`) are provided. This forces users to manually create a wallet and find the ID before they can use the plugin.

There was a TODO in the code to implement automatic wallet creation in this scenario:
```python
# TODO(Kyle): Create Wallet if the wallet ID is not found

## Solution
- Updated `WalletCoinbase` to check for missing `COINBASE_WALLET_ID`.
- If the ID is missing but valid API keys are present, the plugin now automatically creates a new wallet using the Coinbase CDP SDK.
- The new Wallet ID is logged and used immediately, streamlining the setup process.

## Type of Change
- [x] New feature (non-breaking change which adds functionality)

## Testing
- Verified that providing only API keys successfully creates a new wallet and initializes the plugin.
- Verified that providing an existing Wallet ID still works as expected.

### DIFF
--- a/src/inputs/plugins/wallet_coinbase.py
+++ b/src/inputs/plugins/wallet_coinbase.py
@@ -62,20 +62,29 @@ class WalletCoinbase(FuserInput[WalletCoinbaseConfig, List[float]]):
         # TODO(Kyle): Support importing other wallets, following https://docs.cdp.coinbase.com/mpc-wallet/docs/wallets#importing-a-wallet
         API_KEY = os.environ.get("COINBASE_API_KEY")
         API_SECRET = os.environ.get("COINBASE_API_SECRET")
+        api_keys_present = False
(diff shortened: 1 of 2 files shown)


0xbyt4:
Hi @Budalebah, the wallet creation works on init, but when _poll() runs it tries to refresh the wallet using the original env var which is still None so the  first poll after auto-creation will fail. The created wallet's ID needs to be saved back to self.COINBASE_WALLET_ID so subsequent polls can find it. Please   fix it add tests to cover it  and remove todo .

PR #1491, opened 2026-01-15
fix: Copy-paste error in log message - says AI commanda

Vt01nft:
## Summary
bug fix

## Bug
Copy-paste error in log message - says "AI command.action" instead of "Arm command.action"

## Changes
- src/actions/arm_g1/connector/unitree_sdk.py

---

PR #2187, opened 2026-02-10
fix: correct log message in arm_g1 connector

dolepee:
## What
Fixes a copy-paste log message in the arm_g1 Unitree SDK connector.

## Why
The log message says `AI command.action` but this is the **arm** connector, not a generic AI command. Should read `Arm command.action` to match the context (same pattern as other connectors like move/speak).

## Changes
- `src/actions/arm_g1/connector/unitree_sdk.py`: Changed log from `AI command.action` → `Arm command.action`

Same pattern as #2151 (fix: correct log messages in KokoroTTSProvider).

### DIFF
--- a/tests/actions/test_arm_g1_unitree_sdk_connector.py
+++ b/tests/actions/test_arm_g1_unitree_sdk_connector.py
@@ -1,0 +1,52 @@
+import sys
+import types
+import logging
+
+import pytest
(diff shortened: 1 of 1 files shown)


Wanbogang:
@dolepee 
duplicate of #1491

PR #2188, opened 2026-02-10
docs: fix docstring in ActionConnector.connect()

dolepee:
## What
Fixes the docstring in `ActionConnector.connect()` to match the actual parameter name.

## Why
The docstring refers to "input protocol" but the parameter is called `output_interface`. This is misleading for developers implementing custom action connectors.

## Changes
- `src/actions/base.py`: Changed "input protocol" → "output interface" in the `connect()` method docstring (2 lines).

### DIFF
--- a/src/actions/base.py
+++ b/src/actions/base.py
@@ -129,12 +129,12 @@ class ActionConnector(ABC, T.Generic[CT, OT]):
     @abstractmethod
     async def connect(self, output_interface: OT) -> None:
         """
-        Connect the input protocol to the action.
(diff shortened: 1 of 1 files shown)


0xbyt4:
Hi @dolepee , Thanks for contribution !  The parameter actually receives the LLM's parsed input (orchestrator.py calls it `input_interface` before passing it to  `connect()`) , the docstring saying "input protocol" wasn't   entirely wrong, just inconsistent with the parameter name.

PR #2197, opened 2026-02-10
fix(tests): avoid mutable default argument in mock_llm_ask

MuhamadRifansyah:
# Overview
Eliminate mutable default argument in `mock_llm_ask` to prevent shared state across test invocations.

# Change
- Replace `messages=[]` with `messages: Optional[List[Dict[str, str]]] = None`
- Initialize a new list inside the function

# Impact
- No behavior change
- Improves test reliability
- Ruff and Pyright compliant


0xbyt4:
https://github.com/OpenMind/OM1/pull/1447  dublicate.

PR #2239, opened 2026-02-13
Fix VLM camera frame drop handling

### LINKED ISSUE #2207: bug: VLM input plugins crash on dropped camera frames (ret not checked)
**Status:** Fix submitted in PR #2208

## Description

Three VLM input plugins call `cap.read()` but never check the `ret` return value before passing the frame to downstream processing. When the camera drops a frame (returns `ret=False, frame=None`), the plugins crash instead of gracefully skipping.

### Affected files

1. **`src/inputs/plugins/vlm_local_yolo.py:221`** — `ret` is ignored, `frame` (potentially `None`) is passed directly to `self.model.predict()`, which raises a runtime error.

2. **`src/inputs/plugins/vlm_coco_local.py:131`** — `ret` is ignored, `frame` (potentially `None`) is returned and passed to PyTorch model inference.

3. **`src/inputs/plugins/webcam_to_face_emotion.py:89`** — Same pattern, `ret` ignored.

### When this happens

- Virtual camera devices in simulation (Gazebo/Isaac Sim) may not have frames ready immediately at startup
- Physical cameras that temporarily lose connection (USB disconnect, power fluctuation)
- High system load causing the camera driver to drop frames

### Expected behavior

The plugins should check `ret` and return `None` (skip the frame) when `cap.read()` fails, matching the pattern already used in `unitree_realsense_dev_vlm_provider.py:98-109`.

### Steps to reproduce

1. Configure `VLM_Local_YOLO` with a camera index that points to a virtual device not yet streaming
2. Start the OM1 runtime
3. The plugin crashes on the first `_poll()` call when `frame` is `None`

KyokoMai:
## Problem

Fixes #2207

Three VLM camera input modules were not properly validating the return status from `cv2.VideoCapture.read()`, causing them to process invalid frames when cameras drop frames or encounter errors. This could lead to:
- Processing of corrupted/null frame data
- Crashes or undefined behavior in downstream processing
- Silent failures without proper error logging

## Solution

Added proper `ret` status validation immediately after `cap.read()` calls in:

1. **`vlm_local_yolo.py`** (line 221)
2. **`vlm_coco_local.py`** (line 131)
3. **`webcam_to_face_emotion.py`** (line 89)

Each fix follows the same pattern:
```python
ret, frame = self.cap.read()
if not ret:
    logging.warning("Failed to capture frame from camera")
    return None
```

This ensures:
- Invalid frames are never processed
- Clear warning messages are logged for debugging
- Functions return `None` to signal no valid data
- Downstream code handles missing frames gracefully

## Testing

The fix follows the established pattern used elsewhere in the codebase and is consistent with OpenCV best practices for camera frame capture validation.

## Impact

- **Risk**: Low - only adds safety checks
- **Scope**: Camera frame capture validation
- **Behavior**: Graceful handling of frame drops instead of crashes

### DIFF
--- a/src/inputs/plugins/vlm_coco_local.py
+++ b/src/inputs/plugins/vlm_coco_local.py
@@ -31,12 +31,22 @@ class VLM_COCO_LocalConfig(SensorConfig):
 Detection = collections.namedtuple("Detection", "label, bbox, score")
 
 # if working on Mac, please disable continuity camera on your iphone
-# Settings > General > AirPlay & Continuity, and tunr off Continuity
+# Settings > General > AirPlay & Continuity, and turn off Continuity
 
 
 def check_webcam(index_to_check):
     """
     Checks if a webcam is available and returns True if found, False otherwise.
+
+    Parameters
+    ----------
+    index_to_check : int
+        The camera index to check.
+
+    Returns
+    -------
+    bool
+        True if the webcam is available, False otherwise.
     """
     cap = cv2.VideoCapture(index_to_check)  # 0 is the default camera index
     if not cap.isOpened():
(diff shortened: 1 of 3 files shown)


0xbyt4:
The ret check fix  related to issue, but Media_server.mdx changes, docstring additions, and the typo fix are unrelated to frame drop handling and   should be in a separate PR.

PR #2281, opened 2026-02-16
test(providers): add unit tests for singleton decorator

LupaFlex:
## Summary

Added comprehensive unit tests for the singleton decorator in providers which previously had no test coverage.

## Changes

- `tests/providers/test_singleton.py`: Tests for the thread-safe singleton decorator

## Test Coverage

- Singleton returns same instance on multiple calls
- Singleton with constructor arguments (first call wins)
- Reset functionality creates new instance
- Thread safety with 10 concurrent threads
- Singleton class attribute exposure (`_singleton_class`)
- Singleton with keyword arguments
- Multiple singleton classes remain independent

## Testing

- All 7 tests pass locally
- Pre-commit checks pass
- Pyright: 0 errors

## Notes

- Uses direct module loading via `importlib.util` to avoid provider __init__.py import chain
- Thread safety test validates concurrent access with 10 threads

### DIFF
--- a/tests/providers/test_singleton.py
+++ b/tests/providers/test_singleton.py
@@ -1,0 +1,162 @@
+"""Tests for the singleton decorator."""
+
+import importlib.util
+import sys
+import threading
+from unittest.mock import MagicMock
+
+# Mock all problematic modules before any imports
+sys.modules["zenoh"] = MagicMock()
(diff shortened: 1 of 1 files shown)


Wanbogang:
Hi @LupaFlex 
Please add the missing test cases to the existing file tests/providers/test_singleton_provider.py  so don't change the existing name test file

PR #2285, opened 2026-02-16
test(config): skip optional unitree/cyclonedds import failures

dolepee:
## Summary
Split from #2284: this PR only updates  to treat missing optional Unitree/CycloneDDS dependencies as skip/warn conditions, while preserving hard failure for real connector-path/module errors.

## Why
Optional dependency import failures currently produce false negatives in config tests on environments without Unitree SDK deps installed.

## Scope
- Test-only change in 
- No runtime code changes

Follow-up split to keep #2284 runtime fixes narrowly scoped.

### DIFF
--- a/tests/config/test_config.py
+++ b/tests/config/test_config.py
@@ -1,6 +1,9 @@
 import importlib
+import logging
 import os
+import types
 from typing import Optional, Type
+from unittest.mock import patch
 
 import json5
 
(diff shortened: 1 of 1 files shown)


Wanbogang:
Hi @dolepee 
This test should not be skipped. If the connector cannot be imported, the test must fail — that's the intended behavior

PR #2289, opened 2026-02-16
fix: runtime recovery on mode transition/reload failure and wildcard cooldown bypass

0xbyt4:
## Summary

- **Mode transition recovery**: `_on_mode_transition` now falls back to the previous mode when initialization or startup of the new mode fails, instead of leaving orchestrators permanently dead
- **Full state rollback on recovery**: `ModeManager` saves a pre-transition state snapshot before modifying state fields. On failed transition recovery, all state fields (`current_mode`, `previous_mode`, `mode_start_time`, `last_transition_time`, `transition_history`) are reverted and persisted to disk via `revert_transition_state()`
- **Hot-reload recovery**: `_reload_config` now restarts orchestrators with the previous configuration when loading or initializing the new config fails, instead of silently swallowing the error with dead orchestrators
- **Wildcard cooldown bypass**: `_can_transition` now uses the actual current mode for wildcard rule (`from_mode='*'`) cooldown key lookup, matching what `_execute_transition` sets, so cooldown is correctly enforced

## Changes

| File | Change |
|------|--------|
| `src/runtime/cortex.py` | Recovery logic in `_on_mode_transition` and `_reload_config` |
| `src/runtime/manager.py` | Wildcard cooldown key fix in `_can_transition`, snapshot/rollback mechanism in `_execute_transition` and `revert_transition_state()` |
| `tests/runtime/test_cortex.py` | 4 new tests + 1 updated test |
| `tests/runtime/test_manager.py` | 8 new tests |

12 new regression tests added.

## Known limitation

`_execute_transition` runs lifecycle hooks (`on_exit`, `on_enter`) before invoking the transition callback. If the callback fails and recovery is triggered, these hook side effects cannot be undone. This is a pre-existing architectural pattern in the manager's transition flow, not introduced by this PR.

### DIFF
--- a/src/runtime/cortex.py
+++ b/src/runtime/cortex.py
@@ -216,7 +216,7 @@ class ModeCortexRuntime:
             try:
                 await self._initialize_mode(from_mode)
                 await self._start_orchestrators()
-                self.mode_manager.state.current_mode = from_mode
+                self.mode_manager.revert_transition_state()
                 logging.info(f"Successfully recovered to mode: {from_mode}")
             except Exception as recovery_error:
                 logging.critical(
--- a/src/runtime/manager.py
+++ b/src/runtime/manager.py
@@ -77,6 +77,7 @@ class ModeManager:
         self._main_event_loop: Optional[asyncio.AbstractEventLoop] = None
         self._transition_lock = asyncio.Lock()
         self._is_transitioning = False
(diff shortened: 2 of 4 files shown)


0xbyt4:
Splitting into 3 separate PRs for independent review and merge

PR #2298, opened 2026-02-16
fix: correct greeting conversation state logic bugs

wtoqxy0000:
## Summary
- **Bug 1:** `reset_state()` in `greeting_conversation_state_provider.py` ignores its `initial_state` parameter — line 585 unconditionally overwrites `self.current_state` with `CONVERSING`, making the parameter useless. The caller `approaching_person.py` passes `ENGAGING` but it never takes effect.
- **Bug 2:** `connect()` in both `greeting_conversation_elevenlabs.py` and `greeting_conversation_kokoro.py` compares `response.get("current_state")` against `ConversationState.FINISHED` (enum object), but `process_conversation()` returns `self.current_state.value` (string `"finished"`). The comparison is always `False`, so the context update is never sent from `connect()`. The `tick()` method in the same files already correctly uses `.value`.

## Changes
| File | Change |
|------|--------|
| `src/providers/greeting_conversation_state_provider.py:585` | Remove duplicate `self.current_state = ConversationState.CONVERSING` |
| `src/actions/.../greeting_conversation_elevenlabs.py:135` | `ConversationState.FINISHED` → `ConversationState.FINISHED.value` |
| `src/actions/.../greeting_conversation_kokoro.py:149` | `ConversationState.FINISHED` → `ConversationState.FINISHED.value` |
| 3 test files | Update assertions to match corrected behavior |

## Test plan
- [x] All 59 greeting_conversation tests pass locally
- [x] pre-commit hooks pass

### DIFF
--- a/config/greeting_local.json5
+++ b/config/greeting_local.json5
@@ -16,7 +16,7 @@
     config: {
       agent_name: "Bits",
       history_length: 1,
-      base_url: "http://192.168.0.83:8860/v1",
+      base_url: "http://omr2.local:8860/v1",
       model: "nvidia/nemotron-3-nano",
     },
   },
@@ -59,7 +59,7 @@
           connector: "greeting_conversation_kokoro",
           config: {
             model_id: "mlx-community/Kokoro-82M-bf16",
-            base_url: "http://192.168.0.83:8880/v1",
+            base_url: "http://omr2.local:8880/v1",
           },
         },
       ],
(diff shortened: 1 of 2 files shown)


0xbyt4:
1. fix is incomplete: removing line 585 exposes that _determine_next_state() has no ENGAGING handler  maybe  robot will get stuck  ?

wtoqxy0000:
> 1. fix is incomplete: removing line 585 exposes that _determine_next_state() has no ENGAGING handler  maybe  robot will get stuck  ?

Thank u maintainer. I'll check it.

wtoqxy0000:
Fixed. Added an `ENGAGING -> CONVERSING` transition in `_determine_next_state()` so the state machine won't get stuck when `reset_state(ENGAGING)` is called.

```python
# From ENGAGING to CONVERSING
if self.current_state == ConversationState.ENGAGING:
    return ConversationState.CONVERSING
```

This mirrors the existing state flow: ENGAGING is a transient state that should always advance to CONVERSING on the next tick.

PR #2320, opened 2026-02-17
enforce Simulator ABC contract and reject abstract plugins

aydnOktay:
This PR makes Simulator an abstract base class to enforce implementation of sim() and tick() and prevent silent no-op behavior. It also updates the simulator plugin loader to detect and reject abstract simulator subclasses with a clear error message, and adjusts simulator tests to use concrete dummy simulators while adding coverage for the new abstract-class rejection path.


aydnOktay:
@openminddev  @0xbyt4

PR #2347, opened 2026-02-18
test(actions): add interface tests for move_go2_action action

LupaFlex:
## summary

added interface tests for the move_go2_action action module


LupaFlex:
why is the closed

Wanbogang:
please stop making tests that only contain enums and not free strings

PR #2348, opened 2026-02-18
fix: correct backward movement comparison in TurtleBot MoveZenohConnector

dolepee:
## Summary
Fixes backward movement distance comparison logic in .

### Problem
 can be negative for backward movement. Existing comparisons used raw , which causes incorrect stop/overshoot behavior when moving backward.

### Fix
- Normalize comparisons using  for distance-based checks.
- Normalize remaining distance calculation to .

### Why this matters
Prevents incorrect control logic and potential non-termination/overshoot when target displacement is negative.

## Test plan
- Run forward move command () and verify unchanged behavior.
- Run backward move command () and verify proper stop/overshoot handling.
- Verify connector exits when .

### DIFF
--- a/src/actions/move_turtle/connector/zenoh.py
+++ b/src/actions/move_turtle/connector/zenoh.py
@@ -387,7 +387,7 @@ class MoveZenohConnector(ActionConnector[MoveZenohConfig, MoveInput]):
                 distance_traveled = math.sqrt(
                     (self.odom.x - s_x) ** 2 + (self.odom.y - s_y) ** 2
                 )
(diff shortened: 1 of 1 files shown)


0xbyt4:
Hi @dolepee , Thanks for contributions!  PRs #2335, #2336, #2337, #2348 all carry the same CI workflow changes (test ignores, submodule removal, fork guards) duplicated across four unrelated fix branches , submit each fix as its own clean PR; #2308 adds a warning to a method that has zero callers in the codebase.

dolepee:
Thanks for reviewing, I'll work on it.

PR #2350, opened 2026-02-18
test(actions): add interface tests for move_serial_arduino action

LupaFlex:
## summary

added interface tests for the move_serial_arduino action module.

## test coverage

* movementAction enum values and count (4 actions)
* moveInput creation with all action types
* move interface with input/output handling
* 7 tests pass locally, pre-commit and pyright pass


Wanbogang:
@LupaFlex 

Interface only contains pure data definitions (enum and dataclasses) with no logic to test. All enum values and ArmInput usage are already implicitly covered in the connector tests.

PR #2351, opened 2026-02-18
test(actions): add interface tests for move_to_peer action

LupaFlex:
added interface tests for the move_to_peer action module


Wanbogang:
@LupaFlex 
please stop making tests that only contain enums

PR #2353, opened 2026-02-18
fix: correct backward movement comparison in TurtleBot MoveZenohConnector

dolepee:
## Summary
This is a clean, single-purpose PR containing only the TurtleBot backward movement fix.

## Problem
 can be negative for backward movement. Comparing  directly against raw  causes incorrect movement control for reverse targets.

## Fix
- Normalize movement comparisons with .
- Normalize remaining-distance calculation with .

## Files changed
- 

## Why this matters
Prevents incorrect stop/overshoot behavior and improves reliability of backward motion commands.

## Test plan
- Verify forward movement () unchanged.
- Verify backward movement () now advances/stops correctly.
- Verify connector exits when .


dolepee:
Closing this one and replacing with a clean single-file PR to avoid duplicated CI/workflow changes. Please review #2354 (same fix, clean scope).

PR #2355, opened 2026-02-18
fix: start location provider in navigate_location connectors

dolepee:
## Summary
Start the location provider thread in both Unitree navigate connectors.

## Problem
Both connectors instantiate a locations provider but never call `.start()`. Without starting the provider loop, location cache refresh does not run and `get_location()` can return empty/stale results.

## Fix
Add `self.location_provider.start()` in:
- `src/actions/navigate_location/connector/unitree_g1_nav.py`
- `src/actions/navigate_location/connector/unitree_go2_nav.py`

## Scope
Single-purpose PR (no CI/workflow or unrelated file changes).

## Test plan
- Start each connector and verify provider thread starts.
- Request known location labels and verify `get_location()` resolves correctly.
- Verify navigation goal publication still works unchanged.


Wanbogang (approved):

Prachi1615:
Thanks for the contribution! Closing this PR as it targets the old Python code, which has now been migrated to Go.

PR #2358, opened 2026-02-18
fix: guard failed camera frame reads in VLM COCO local plugin

dolepee:
## Summary
Guard webcam frame reads in COCO local plugin to avoid propagating invalid frames.

## Problem
Camera reads can intermittently fail. Returning an invalid frame can cause downstream errors.

## Fix
In `src/inputs/plugins/vlm_coco_local.py`:
- Return frame only when `ret` is true.
- Otherwise return `None` gracefully.

## Scope
Single-file, single-purpose fix.


Wanbogang (approved):

Prachi1615:
Thanks for the PR! 
The project has since migrated from Python to Go, so this PR is no longer applicable. Closing it for now.

PR #2377, opened 2026-02-19
fix(hooks): log ignored lifecycle hook failures as warnings

aydnOktay:
This PR improves observability of the new lifecycle hook infrastructure by logging a warning whenever a hook fails but is configured with **on_failure="ignore"**, instead of silently passing. The change updates **execute_lifecycle_hooks** to emit a clear warning message and extends the existing tests to assert that ignored failures are both reported and still cause the function to return False, keeping behavior consistent while making debugging easier.

### DIFF
--- a/config/greeting_local.json5
+++ b/config/greeting_local.json5
@@ -21,6 +21,12 @@
     },
   },
 
+  knowledge_base: {
+    knowledge_base_name: "demo",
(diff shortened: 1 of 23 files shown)


aydnOktay:
@openminddev  @0xbyt4

aydnOktay:
@openminddev

PR #2395, opened 2026-02-20
fix: add HEALTHCHECK instruction to Dockerfile

DasRasyo:
## Summary
- Add `HEALTHCHECK` instruction to the Dockerfile
- The container currently has no health check mechanism, so orchestration systems (Docker Swarm, Kubernetes, Watchtower) cannot detect if the service becomes unresponsive
- Uses a lightweight Python probe with a 120s start period to account for network and audio initialization in the entrypoint

## Test plan
- [ ] `docker build -t om1-test .` completes successfully
- [ ] `docker inspect om1-test | grep -A5 Healthcheck` shows the configured health check
- [ ] Running container reports `healthy` status via `docker ps`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -93,5 +93,8 @@ RUN echo '#!/bin/bash' > /entrypoint.sh && \
     echo 'fi' >> /entrypoint.sh && \
     chmod +x /entrypoint.sh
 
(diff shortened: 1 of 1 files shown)


Wanbogang:
@DasRasyo 
the container already has restart: unless-stopped

PR #2397, opened 2026-02-20
fix: split redundant (JSONDecodeError, Exception) into separate except blocks

johnh4098:
## Type of Change
- [x] Bug fix

## Changes

`except (json.JSONDecodeError, Exception)` is used in two files:
- `src/runtime/manager.py` (line 793)
- `src/backgrounds/plugins/unitree_go2_frontier_exploration.py` (line 54)

`Exception` is a superclass of `json.JSONDecodeError`, making the specific type in the tuple meaningless — this is equivalent to bare `except Exception`. This means unrelated errors (e.g. `AttributeError`, `TypeError`) are caught and logged as JSON errors, hiding real bugs.

Split into two separate `except` blocks so each exception type is handled independently. Behavior and error messages are preserved.

## Checklist
- [x] Self-review completed
- [x] Local testing completed

## Impact
No behavioral change for JSON decode errors. Non-JSON exceptions now propagate their correct type information instead of being masked as JSON errors.

### DIFF
--- a/src/backgrounds/plugins/unitree_go2_frontier_exploration.py
+++ b/src/backgrounds/plugins/unitree_go2_frontier_exploration.py
@@ -51,7 +51,10 @@ class UnitreeGo2FrontierExploration(Background[UnitreeGo2FrontierExplorationConf
 
         try:
             context_aware_text = json.loads(context_aware_text)
-        except (json.JSONDecodeError, Exception) as e:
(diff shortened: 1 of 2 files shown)


Wanbogang:
@johnh4098 
Both except blocks do the same thing — no reason to split them.

PR #2412, opened 2026-02-22
fix: correct Queue truthiness check in TwitterInput plugin

### LINKED ISSUE #2411: self.message_buffer is a Queue object. In Python, a Queue instance is always truthy regardless of whether it's empty, so the check if self.message_buffer: never evaluates to False.
**Describe the bug**
`self.message_buffer` is a `Queue` object. In Python, a Queue instance 
is always truthy regardless of whether it's empty, so the check 
`if self.message_buffer:` never evaluates to False.

**To Reproduce**
- Run any config that uses the TwitterInput plugin
- The buffer check on line ~13777 of twitter.py always passes
- get_nowait() is called even on an empty queue, relying on the 
  Empty exception as control flow rather than an intentional guard

**Expected behavior**
The code should check `if not self.message_buffer.empty():` to 
properly guard the get_nowait() call.

**System Information (please complete the following information):**
 - **Operating System:** [Windows 10]
 - **Python Version:** [e.g. v3.11]

Precious-Neche:
## Overview

`self.message_buffer` is a `Queue` object. In Python, a Queue instance 
is always truthy regardless of whether it is empty, meaning the guard 
`if self.message_buffer:` never prevents the `get_nowait()` call on an 
empty queue. The code currently relies on catching the `Empty` exception 
as control flow rather than properly checking first.

## Type of change
Fix

## Changes
Fixed a bug in src/inputs/plugins/twitter.py where self.message_buffer (a queue.Queue object) was being used as a boolean condition with if self.message_buffer:. In Python, a Queue instance is always truthy regardless of whether it is empty or not, meaning this check never evaluates to False and provides no actual guard.
Replaced if self.message_buffer: with if not self.message_buffer.empty(): so the condition correctly reflects whether the buffer has messages before attempting to call get_nowait().

## Checklist

- [ ] Code complies with style guidelines (check)
- [ ] Self-review completed (check)
- [ ] Documentation updated (if applicable)
- [ ] Local testing completed(check)
- [ ] Tests added/updated (if applicable)

## Impact
Low risk, isolated change. The fix corrects a silent logic error where the empty queue check was never working as intended. The behavior of the code is unchanged in practice since the Empty exception was already being caught below, but the fix makes the guard explicit and correct rather than relying on exception handling as control flow.

## Additional Information
Closes #2411

### DIFF
--- a/src/inputs/plugins/twitter.py
+++ b/src/inputs/plugins/twitter.py
@@ -118,7 +118,7 @@ class TwitterInput(FuserInput[TwitterSensorConfig, Optional[str]]):
         if raw_input:
             self.message_buffer.put_nowait(raw_input)
 
-        if self.message_buffer:
+        if not self.message_buffer.empty():
             try:
                 message = self.message_buffer.get_nowait()
                 if message:


0xbyt4:
Hi @Precious-Neche  Thanks for contribution! The behavior is unchanged  Empty is already caught on line 128, so the if check was never the actual guard here.

PR #2413, opened 2026-02-23
refactor(twitter): simplify TwitterInput following NewsInput pattern

Wanbogang:
## Summary
Refactors `TwitterInput` to follow the same pattern as `NewsInput`, 
fixing the root cause behind #2411 and several other issues found 
during analysis.

## Problems Found
- `if self.message_buffer:` always evaluates to `True` (reported in #2411)
- `listen()` was overriding base class unnecessarily
- Context was only fetched once at `start()`, never refreshed during runtime
- Data flowed through Queue in circles: `_poll()` → `raw_to_text()` → back to Queue
- `start()`, `initialize_with_query()`, `_query_context()` were dead code
- No rate limiting on API calls (called every 0.5s regardless of need)
- `io_provider.add_input()` was missing, inconsistent with other plugins
- `self.buffer` and `self.context` were redundant and confusing

## Changes
- Remove `Queue`, `message_buffer`, `start()`, `listen()` override,
  `_query_context()`, `initialize_with_query()`
- Add `poll_interval` to `TwitterSensorConfig` (default: 60s)
- `_poll()` fetches API directly with interval guard
- `formatted_latest_buffer()` now calls `io_provider.add_input()`


oholiab0818:
good

PR #2432, opened 2026-02-28
fix(orchestrator): handle exceptions in _promise_action and flush_promises

Wanbogang:
## Problem

Two silent failure points existed in `ActionOrchestrator`:

1. `_promise_action()` had no exception handling around `input_type(**converted_params)` and `connector.connect()`. Any crash here — invalid enum value from LLM, missing required parameter, or hardware/network failure — would be swallowed silently by `asyncio.wait()` in `flush_promises()`.

2. `flush_promises()` never retrieved exceptions from completed tasks. Since `asyncio.wait()` does not re-raise task exceptions, failures disappeared entirely with no log output and no trace.

## Fix

- Wrap `input_type(**converted_params)` in try/except with `logging.exception()` — handles invalid LLM output such as unknown enum values or missing required fields
- Wrap `connector.connect()` in try/except with `logging.exception()` — handles hardware disconnect, network timeout, or any runtime failure in the connector
- Add exception retrieval loop in `flush_promises()` so failed tasks are logged before being discarded


Prachi1615:
@Wanbogang Thanks for the PR! 
The project has since migrated from Python to Go, so this PR is no longer applicable. Closing it for now.

PR #2434, opened 2026-02-28
fix: consistent score calculation in FAISSRetriever

0xbyt4:
## Summary
- `search()` returned raw FAISS distance as score while `batch_search()` used `1/(1+dist)` normalization
- Same distance value produced different scores: e.g., dist=0.1 gave score 0.1 via `query()` but 0.909 via `query_batch()`
- Aligned `search()` to use the same `1/(1+dist)` formula as `batch_search()`


YuchengZhou821:
@0xbyt4  Hi, this is not the correct way. Since we are using inner product in faiss file, therefore `score = dist` is correct.

Prachi1615:
Hi @0xbyt4 
Closing this PR as it targets the old Python code, which has now been migrated to Go.

PR #2664, opened 2026-08-04
fix(ws): start read and write loops for non-reconnecting clients

### LINKED ISSUE #2665: [BUG] WebSocket client does not send or receive when Reconnect is disabled
**Describe the bug**

When a WebSocket client is created with `Reconnect: false`, `Connect()` returns successfully after the initial dial, but the client cannot send or receive messages.

`Client.Connect()` currently returns directly after `dial()` when reconnecting
is disabled:

```go
if !c.cfg.Reconnect {
    return c.dial()
}
```

The sendLoop and readLoop goroutines are only started in the
Reconnect: true branch. As a result:

- Connect() returns nil;
- Send() may also return nil because it only enqueues data into sendCh;
- no goroutine consumes sendCh, so the data is never written;
- no read loop is running, so onMessage is never called.

This also conflicts with the Connect() comment, which states that it starts
the send and read goroutines.

**To Reproduce**
List the steps to replicate the bug:

1. Start a WebSocket echo server.

2. Create the client with:
```go
client := ws.New(ws.Config{
    URL:       url,
    Reconnect: false,
}, log, onMessage)
```

3. Call client.Connect().

4. Call client.Send([]byte("ping")).

5. Observe that Send() may return successfully, but the server does not receive the message and the echo callback is not invoked.

**Expected behavior**

After a successful initial connection, the client should be able to send and receive messages regardless of whether reconnecting is enabled.

Reconnect should control retry behavior after connection failures, not whether the normal send and read loops are started.

**Screenshots**
If relevant, attach screenshots to illustrate the problem.

**System Information (please complete the following information):**
 - **Operating System:** [e.g., macOS with M-series chip]
 - **Python Version:** [e.g. v3.10]

**Additional context**

Start sendLoop and readLoop after every successful initial dial, including when Reconnect is disabled.

grandpig:
## Overview

Fix WebSocket clients configured with `Reconnect: false` so they can send and receive messages after a successful initial connection.

When `Reconnect` is disabled, `Client.Connect()` returns immediately after a successful dial without starting the send and read goroutines.

Consequently, `Send()` can enqueue data successfully, but no goroutine consumes the queue, and incoming messages are never delivered to `onMessage`.

Fixes https://github.com/OpenMind/OM1/issues/2665

## Type of change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation improvement
- [ ] Bounty issue submission
- [ ] Other:

## Changes
[Detail the changes you have made in this pull request. Include any new features, bug fixes, or improvements.]

## Checklist

- [x] Self-review completed
- [ ] Documentation updated (if applicable)
- [x] Local testing completed
- [ ] Tests added/updated (if applicable)

## Impact
[Explain the impact of your changes. Include any potential risks or side effects that reviewers should be aware of.]

- Start `sendLoop` and `readLoop` after every successful initial dial.
- Keep `Reconnect` responsible for retrying failed initial connections and
  reconnecting after subsequent connection errors.
- Add `TestConnectSendReceiveWithoutReconnect` as a regression test.

## Additional Information
[Include any additional information that may be relevant to reviewers. This could include links to related issues or pull requests, references to documentation, or other context that may help reviewers understand your changes.]

### DIFF
--- a/internal/ws/client.go
+++ b/internal/ws/client.go
@@ -60,7 +60,13 @@ func New(cfg Config, log *zap.Logger, onMessage func(messageType int, data []byt
 // Connect dials the server and starts the send/read goroutines.
 func (c *Client) Connect() error {
 	if !c.cfg.Reconnect {
-		return c.dial()
+		if err := c.dial(); err != nil {
+			return err
+		}
+		c.wg.Add(2)
+		go c.sendLoop()
+		go c.readLoop()
+		return nil
 	}
 
 	for {
--- a/internal/ws/client_test.go
+++ b/internal/ws/client_test.go
@@ -68,6 +68,25 @@ func TestConnectSendReceive(t *testing.T) {
 	}
 }
 
+func TestConnectSendReceiveWithoutReconnect(t *testing.T) {
+	url := echoServer(t)
+
+	received := make(chan []byte, 1)
+	c := New(Config{URL: url, Reconnect: false}, zap.NewNop(), func(_ int, data []byte) {
+		received <- append([]byte(nil), data...)
+	})
+	require.NoError(t, c.Connect())
+	t.Cleanup(c.Close)
+
+	require.NoError(t, c.Send([]byte("ping")))
+	select {
+	case got := <-received:
+		require.Equal(t, []byte("ping"), got, "the server echoes the message back to onMessage")
+	case <-time.After(2 * time.Second):
+		t.Fatal("did not receive echoed message")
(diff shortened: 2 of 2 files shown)


grandpig:
@kyle-openmind This is a small WebSocket client fix: non-reconnecting clients now start the same send/read loops after a successful initial dial, with an end-to-end echo regression test. Would you mind taking a look when you have a chance? Thanks!

PR #2677, opened 2026-08-28
[test] PR Triage smoke test — safe to close

25marcusb:
Throwaway PR to exercise the real `pull_request_target` trigger on the PR Triage workflow (`.github/workflows/pr-triage.yml`) against `worktree-om1-pr-triage-workflow`. Opened by an OpenMind org member, so it should hit the org-membership skip path and post nothing. Safe to close/delete after verifying the run.


25marcusb:
Closing — turns out pull_request_target can't fire from a non-default base branch at all, so this test was never going to trigger. Cleaning up.

PR #2680, opened 2026-08-28
Add PR Triage: automated flagging of low-value external PRs

25marcusb:
## Overview

Adds a new **PR Triage** GitHub Action (`.github/workflows/pr-triage.yml`) that automatically flags and closes low-value open PRs from non-OpenMind-org contributors, so maintainers spend less time triaging spam/trivial/off-topic submissions.

## How it works

- **Trigger:** `pull_request_target` (opened/synchronize/reopened) for real-time triage as PRs come in, plus a daily 13:00 UTC cron as a fallback sweep of every open PR (catches anything the event-based path missed).
- **Gate:** checks OpenMind org membership (via `ORG_READ_TOKEN`) and skips org members entirely; skips PRs already triaged at their current commit (idempotency marker comment).
- **Judgment:** Gemini CLI reads `README.md`/`CONTRIBUTING.md` for grounding, then evaluates each candidate PR against three criteria only — IRRELEVANT, TRIVIAL, or predominantly PYTHON files.
- **Action:** if flagged, posts a `Request changes` review explaining exactly why (citing specific evidence), adds a `needs-human-review` label, then **closes the PR**. The review body tells the contributor it was closed automatically and that a maintainer can reopen it if it was flagged in error. If not flagged, does nothing. **Never** approves or merges — closing is the only autonomous action taken, and a human can always reopen a closure that was a mistake.

## Required secrets

- `ORG_READ_TOKEN` — GitHub PAT with `read:org` scope, for the org-membership check. **Now live** in repo secrets.
- `GEMINI_API_KEY` — already present in this repo (used by the existing Docs Sync workflow).

## Testing

Fully verified end-to-end, twice, against real PRs from a non-org burner account (both safely closed/cleaned up, not merged):

1. **First pass** (review/label/marker only, pre-auto-close): caught and fixed a real bug — Gemini CLI's `run_shell_command` tool spawns `gh` in a subprocess that doesn't inherit `GH_TOKEN` from the step's environment, so every `gh` call from inside Gemini's own tool-use loop failed authentication even though the surrounding script's `gh` calls worked fine. Fixed by authenticating `gh` to disk (`gh auth login --with-token`) before invoking `gemini`.
2. **Second pass** (full pipeline including auto-close): confirmed the close step fires correctly and in the right order — review posted (with the new "closed automatically, reopen if this was a mistake" note) → label applied → PR actually closed → idempotency marker comment. Also caught and fixed a smaller gap: `gh label create ... || true` silently no-ops when the label already exists, so a label-description update can go live in the workflow file without ever taking effect on a repo where the label predates the change. Fixed by falling back to `gh label edit` instead of a bare no-op, and manually corrected the already-stale label in this repo.

**Not yet exercised for real:** the daily cron/batch fallback path itself (shares the same verified downstream logic as the tested single-PR path, minus the initial `gh pr list` call, which was already proven working).

**Policy note:** auto-closing is a deliberate escalation from the original design, which kept the bot out of the accept/reject decision entirely. Worth a second look before merge given it acts autonomously on real contributors' PRs.

**Note on commit history:** this branch's commit log includes exploratory/diagnostic commits from testing (temporary trigger overrides, diagnostic re-runs, etc.) — the net diff is a single clean file; happy to squash the history if preferred.


.github/scripts/gh-triage-wrapper.sh
```diff
@@ -0,0 +1,99 @@
...
+#
+# Deliberately NOT configured via environment variables: this workflow
+# already hit one confirmed case (see pr-triage.yml's gh-auth comment) of
+# env vars set for this step not reliably reaching subprocesses spawned by
+# Gemini CLI's shell tool. So the real gh path is baked into this file as a
+# literal string at install time (see the sed substitution in pr-triage.yml
+# — REAL_GH_PATH_PLACEHOLDER below is replaced there, never left as-is),
+# and the allowed PR number is read from a plain file in the current
+# directory instead, the same way this workflow already shares
+# .pr-triage-candidates.tsv with Gemini's shell commands.
```
openminddev: Please remove the comments

.github/workflows/pr-triage.yml
```diff
@@ -0,0 +1,267 @@
...
+#
+# Prompt-injection containment: the triage step hands Gemini (running with
+# no per-command approval) untrusted PR content — the external contributor's
+# own title/body/diff — while it holds a gh token that can write to any PR
+# or issue in the repo. Rather than relying on the prompt text to keep the
+# model in its lane, that step replaces `gh` itself with a validating
+# wrapper (.github/scripts/gh-triage-wrapper.sh) that only allows the exact
+# subcommands this workflow needs, and only against the one PR each Gemini
+# invocation was launched for. See that script and the step below for
+# details.
```
openminddev: Please remove the comments.

openminddev (approved):
Please remove comments.

openminddev (approved):

PR #2683, opened 2026-08-31
Include LLM text content in trace output

WenjinFu:
## Summary
- `traceOutput` in `internal/runtime/config.go` only ever serialized `response.ToolCalls`, so `llm_output` in `traces/*.jsonl` was `[]` for every conversational (non-tool-call) turn — the model's actual text reply was never recorded.
- Now includes a `{"type":"text","value":<TextContent>}` entry ahead of the tool-call entries whenever `response.TextContent` is non-empty.

Found while investigating why every line in a robot's `tracer_2026-08-31.jsonl` had `"llm_output":[]` despite active conversation.

## Test plan
- [x] `go test ./internal/runtime/... ./internal/providers/...` passes
- [x] `go vet ./internal/runtime/... ./internal/providers/...` passes
- [x] Added `TestTraceOutputIncludesText` and `TestTraceOutputEmptyTextOmitted` covering the new behavior

https://claude.ai/code/session_01WtDnirL8kmxDz482UjMtvF

### DIFF
--- a/internal/runtime/config.go
+++ b/internal/runtime/config.go
@@ -229,9 +229,15 @@ func toolCallsToMaps(toolCalls []llm.ToolCall) []map[string]any {
 	return result
 }
 
-// traceOutput converts an LLM response's tool calls into a slice of maps for structured tracing.
+// traceOutput converts an LLM response's text and tool calls into a slice of maps for structured tracing.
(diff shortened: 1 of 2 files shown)


25marcusb:
Closing to reopen under the correct GitHub account.

PR #2684, opened 2026-08-31
Include LLM text content in trace output

25marcusb:
## Summary
- `traceOutput` in `internal/runtime/config.go` only ever serialized `response.ToolCalls`, so `llm_output` in `traces/*.jsonl` was `[]` for every conversational (non-tool-call) turn — the model's actual text reply was never recorded.
- Now includes a `{"type":"text","value":<TextContent>}` entry ahead of the tool-call entries whenever `response.TextContent` is non-empty.

Found while investigating why every line in a robot's `tracer_2026-08-31.jsonl` had `"llm_output":[]` despite active conversation. Verified live: built and deployed this branch's image on the robot (`greeting_conversation_TRMI` config), confirmed new trace lines now include the `"type":"text"` entry.

(Supersedes #2683, closed to reopen under the correct account.)

## Test plan
- [x] `go test ./internal/runtime/... ./internal/providers/...` passes
- [x] `go vet ./internal/runtime/... ./internal/providers/...` passes
- [x] Added `TestTraceOutputIncludesText` and `TestTraceOutputEmptyTextOmitted` covering the new behavior
- [x] Verified live on a deployed robot: real conversational trace lines now show `"type":"text"` entries in `llm_output`

https://claude.ai/code/session_01WtDnirL8kmxDz482UjMtvF


openminddev:
Could you please remove the trace export? We can enable it as the default. It will be a part of the trace. Thanks.

